PR #1, opened 2020-03-12
Correct bugfix in Metadata-standard-names.md

climbfuji:
Correct bugfix in Metadata-standard-names.md (replace _loop_extent with _dimension)

### DIFF
--- a/Metadata-standard-names.md
+++ b/Metadata-standard-names.md
@@ -20,7 +20,7 @@ Dimension standard names come in sets of six related standard names for each dim
 ```
 Note that the cap generator may substitute among standard names in this category in order to properly call suite parts and individual schemes. In the substitutions below, the name on the left is the standard_name in the dimensions field of the caller while the name(s) on the right is (are) the standard name(s) of the callee (in the form used in the subroutine call).
 ```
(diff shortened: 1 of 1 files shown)


gold2718:
Two comments:

1. This is not a bug, it happens when the host model metadata has `horizontal_dimension` but the scheme has `horizontal_loop_extent` (i.e., all the time for `xx_run`` routines).
2. The markdown file is generated from the XML file, please edit the XML and follow the instructions in README.md to regenerate the markdown.

climbfuji:
So shouldn't the host model provide both? Or are we ok with having only horizontal_dimension and use it for the _run routines when horizontal_loop_extend does not exist?

gold2718:
The metadata for a host model field (say temperature) has one entry. That entry uses `horizontal_dimension` because that is the size of the data on the host model.
At run time, the host model will either pass the entire array, or will pass part of it in a loop (which may be threaded). In either case, the schemes interpret whatever the host model is sending as `horizontal_loop_extent`. That is why capgen can make that substitution during the analysis of `xx_run` scheme interfaces.

climbfuji:
Got it, thanks for the explanation.

PR #6, opened 2020-06-18
python2 needs to be used for .xml -> .md

davegill:
The generic `python` will usually be interpreted as python3 now-a-days. 

Just make the README explicit.
```
python2 write_standard_name_table.py standard_names.xml
```

modified:   README.md


gold2718:
Since python2 is EOL, it will be much better to just ensure python 3 compatibility.

PR #14, opened 2021-05-18
Rules for creating CCPP standard names

ligiabernardet:
Initial import of a file containing information about the rules for CCPP standard names. The information includes references to the CF convention, a list of qualifiers (extending the CF convention), a list of generic names (with corresponding units) that are used when creating standard names, and a list of acronyms, abbreviations, and aliases.


StandardNamesRules.rst
```diff
@@ -0,0 +1,513 @@
...
+| due_to_overturning
+| due_to_shallow_convection
+| due_to_shortwave_heating
+| due_to_thermodynamics
+| due_to_background
+| **due_to_subgrid_scale_vertical_mixing**
+| **due_to_convective_microphysics**
+| **due_to_model_physics**
+| **due_to_convective_gravity_wave_drag**
+| **due_to_shoc**
```
climbfuji, line 203: Is this too specific?

StandardNamesRules.rst
```diff
@@ -0,0 +1,513 @@
...
++---------------------+---------------------------------------------------------+
+| min                 | minimum                                                 |
++---------------------+---------------------------------------------------------+
+| nir                 | near-infrared part of the EM spectrum (radiation)       |
++---------------------+---------------------------------------------------------+
+| stp                 | standard temperature (0 degC) and pressure (101325 Pa)  |
++---------------------+---------------------------------------------------------+
+| tke                 | turbulent kinetic energy                                |
++---------------------+---------------------------------------------------------+
+| TKE                 | turbulent kinetic energy                                |
```
climbfuji, line 500: Aren't standard names case insensitive?

StandardNamesRules.rst
```diff
@@ -0,0 +1,513 @@
...
+   in_stratosphere), process (e.g. due_to_deep convection), or condition (e.g.,
+   assuming_clear_sky). The order defined by the CF rules should be observed. These
+   qualifications do not change the units of the quantity.
+
+   All of the following phrases in brackets are optional. The words in ``this font``
+   appear explicitly as stated, while the words in *italic* indicate other
+   words or phrases to be substituted. The new standard name is constructed by
+   joining the base standard name to the qualifiers using underscores.
+
+   [*surface*] [*component*] standard_name [``at`` *surface*] [``in`` *medium*]
```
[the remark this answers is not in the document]
ligiabernardet: This is part of the [CF convection](http://cfconventions.org/Data/cf-standard-names/docs/guidelines.html#surface). The explanation they provide is : "In the standard name, some surfaces are named by single words which are placed at the start: toa (top of atmosphere), tropopause, surface. Other surfaces are named by multi-word phrases put after at: at_adiabatic_condensation_level, at_cloud_top, at_convective_cloud_top, at_cloud_base, at_convective_cloud_base, at_freezing_level, at_ground_level, at_maximum_wind_speed_level, at_sea_floor, at_sea_ice_base, at_sea_level, at_top_of_atmosphere_boundary_layer, at_top_of_atmosphere_model, at_top_of_dry_convection."

If desirable, I can add more details to the CCPP rules, so it is clearer when to put the surface at the beginning of the name versus at the end.

StandardNamesRules.rst
ligiabernardet: Yes, so I think I can remove the upper case aliases.

StandardNamesRules.rst
ligiabernardet: All these names are already in use in https://github.com/ESCOMP/CCPPStandardNames/pull/13. In other words, they are needed in the current CCPP. Therefore, I think we need them.

StandardNamesRules.rst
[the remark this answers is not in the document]
climbfuji: Thanks for the explanation, I'll leave it up to you whether to provide further details in the CCPP rules or not.

StandardNamesRules.rst
ligiabernardet: Fixed

StandardNamesRules.rst
```diff
@@ -0,0 +1,507 @@
...
+
+#. Variables are current and instantaneous unless specified. Variables that are not
+   current (e.g., previous timestep) or non-instantaneous (e.g., accumulated values)
+   should have qualifiers in the standard name to describe what they represent.
+
+#. By default (when not specified otherwise), variables are grid means or centers
+   (defined by the host). If a variable is defined at a different physical location,
+   a qualifier should be used to denote this. For example, for variables
+   representing quantities at the interface between grid cells vertically,
+   use at_interface.
```
gold2718, line 57: I think this rule may be in conflict with CF which sometimes seems to make different assumptions (not always self-consistent). It would be great to have a "get out of rule free" card that says to explicitly use `layer` to resolve ambiguous situations.

cacraigucar (changes requested):

climbfuji (approved):

StandardNamesRules.rst
```diff
@@ -0,0 +1,507 @@
...
+   narrowly-defined context or a variable without the scope-narrowing qualifiers
+   already exists and cannot be reused.
+
+#. Spell out acronyms unless they are obvious to a vast majority of
+   scientists/developers who may come across them. Here is a
+   :ref:`list of currently-used aliases <Aliases>` where either is
+   acceptable.
+
+#. For control-oriented variables, if the variable is a Fortran logical,
+   use flag_for_X. If it is any other data type, use control_for_X.All flags
```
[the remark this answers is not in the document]
ligiabernardet: This is a good suggestion but I do not know how to write a RST file that will display in color. We were using _italics_ to represent placeholders and **bold** to represent qualifiers that do not exist in the CF convention. However, RST does not support nested special characters such as **_bold italic_**. Do you have any suggestion?

StandardNamesRules.rst
```diff
@@ -0,0 +1,507 @@
...
+
+Acronyms, Abbreviations, and Aliases
+====================================
+
++---------------------+---------------------------------------------------------+
+| **Short**           |  **Meaning**                                            |
++=====================+=========================================================+
+| ir                  | infrared                                                |
++---------------------+---------------------------------------------------------+
+| IR                  | infared                                                 |
```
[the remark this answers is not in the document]
ligiabernardet: Agreed, I will do that.

StandardNamesRules.rst
```diff
@@ -0,0 +1,507 @@
...
+| at_freezing_level
+| at_ground_level
+| at_maximum_wind_speed_level
+| at_sea_ice_base
+| at_sea_level
+| at_top_of_atmosphere_boundary_layer
+| at_top_of_atmosphere_model
+| at_top_of_dry_convection
+| **at_interface**
+| **at_surface_adjacent_layer OR at_bottom_layer OR at_lowest_model_layer**
```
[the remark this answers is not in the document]
ligiabernardet: These options are historic and I will remove them. The only remaining qualifier will be at_surface_adjacent_layer. The other qualifiers (at_bottom_layer and at_lowest_model_layer) are not used in the dictionary of Standard Names.

StandardNamesRules.rst
```diff
@@ -0,0 +1,507 @@
...
+| **assuming_shallow**
+| **assuming_deep**
+
+Time
+----
+
+Suffixes
+^^^^^^^^
+
+| **of_new_state OR updated_by_physics**
```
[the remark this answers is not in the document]
ligiabernardet: I will remove updated_by_physics since it is not used in the dictionary of Standard Names The only remaining qualifier will be of_new_state.

StandardNamesRules.rst
```diff
@@ -0,0 +1,507 @@
...
+| **for_current_cubed_sphere_tile**
+| **plus_one**
+| **minus_one**
+| **one_way_coupling_of_X_to_Y**
+| **for_radiation**
+| **for_deep_convection**
+| **for_microphysics**
+| **directory_for_X_source_code**
+| **flag_for_reading_X_from_input**
+| **tunable_parameters[s]_for_X**
```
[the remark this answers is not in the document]
ligiabernardet: One "s" should be removed (I will change this). The qualifier is tunable_parameter[s].
Tunable parameters are coefficients within parameterizations. They can be "tuned" to achieve the desired results in a simulation. They are considered tunable because they do not have a correct known value (a range of values may be acceptable).

StandardNamesRules.rst
```diff
@@ -0,0 +1,507 @@
...
+| **on_dynamics_timestep**
+| **on_radiation_timestep**
+| **on_previous_timestep**
+| **N_timesteps_back**
+
+Computational
+-------------
+
+Prefixes and Suffixes
+^^^^^^^^^^^^^^^^^^^^^
```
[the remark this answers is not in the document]
ligiabernardet: I will separate them

StandardNamesRules.rst
ligiabernardet: The CF is not geared toward numerical modeling, so it does not have subtleties such as layers vs interfaces. So I think we have discretion here to make our own rule. I am not seeing any ambiguous situations at the moment, that is: 1) default is layer, 2) if not layer, a qualifier must be used to denote  where the variable is defined. 
Do you have any suggestion?

StandardNamesRules.rst
```diff
@@ -0,0 +1,507 @@
...
+
+| **real**
```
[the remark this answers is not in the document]
ligiabernardet: Yes, it is referring to a computer _real_ vs _integer_. I will look at those variables and get back to you.

StandardNamesRules.rst
```diff
@@ -0,0 +1,507 @@
...
+   assuming_clear_sky). The order defined by the CF rules should be observed. These
+   qualifications do not change the units of the quantity.
+
+   All of the following phrases in brackets are optional. The words in ``this font``
+   appear explicitly as stated, while the words in *italic* indicate other
+   words or phrases to be substituted. The new standard name is constructed by
+   joining the base standard name to the qualifiers using underscores.
+
+   [*surface*] [*component*] standard_name [``at`` *surface*] [``in`` *medium*]
+   [``due_to`` *process*] [``assuming`` *condition*]
```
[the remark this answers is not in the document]
ligiabernardet: "Transformation" qualifiers always appear the beginning of the standard name, so I will make that clear in the positional hierarchy. The "Computational" qualifiers are more complex, in that sometimes they appear at the beginning, such as _flag_for_, and sometimes at the end, such as _for microphysics_. At this time I propose that we do not add computational qualifiers to the positional hierarchy, and we can revisit the topic in the future.

StandardNamesRules.rst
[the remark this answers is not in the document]
cacraigucar: My knowledge of RST is limited. Before I made the request, I did a google search and found this:  https://stackoverflow.com/questions/4669689/how-to-use-color-in-text-with-restructured-text-rst2html-py-or-how-to-insert-h

If this works, I'd suggest using color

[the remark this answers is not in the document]
ligiabernardet: I  also have been googling. I don't think the suggestion you sent will work because it depends on CCS/HTML to display. 
I will have a commit soon with a proposed format to address this issue. Coming soon.

StandardNamesRules.rst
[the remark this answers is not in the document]
ligiabernardet: I agree that the __real_ qualifier is awkward because this information should be in the _type_ metadata, and not in the standard name. However, today we need this qualifier because the CCPP Physics has the same variable in integer and real form and the CCPP Framework does not distinguish them. So we need the distinction to be in the standard name.
After discussion within the CCPP team at DTC, here is what we suggest for the way forward. We will eliminate this undesirable rule even though we have _real standard names, and we will do follow up development to solve this problem. See https://github.com/NCAR/ccpp-physics/issues/704.

ligiabernardet:
@cacraigucar and @gold2718 I believe I have addressed your comments, pls take a look.

StandardNamesRules.rst
```diff
@@ -0,0 +1,510 @@
...
+Computational
+-------------
+
+Prefixes
+^^^^^^^^
+
+| **for_coupling**
+| **for_chemistry_coupling**
+| **from_coupled_process**
+| **from_wave_model**
```
cacraigucar, line 249: These are suffixes

StandardNamesRules.rst
```diff
@@ -0,0 +1,510 @@
...
+^^^^^^^^
+
+| net
+| upwelling
+| downwelling
+| incoming
+| outgoing
+
+Medium
+------
```
cacraigucar, line 152: There is nothing here (should it be removed)?

StandardNamesRules.rst
```diff
@@ -0,0 +1,510 @@
...
+| **multiplied_by_timestep**
+| **for_current_mpi_rank**
+| **for_current_cubed_sphere_tile**
+| **plus_one**
+| **minus_one**
+| **for_radiation**
+| **for_deep_convection**
+| **for_microphysics**
+| **directory_for** ``_X`` **_source_code**
+| **flag_for_reading** ``_X`` **_from_input**
```
cacraigucar, line 286: These are not suffixes (since they are in the middle).  Should there be a `middle` or `infix` category for these two.  (Note I didn't know the word `infix` before I googled it).

cacraigucar (changes requested):

StandardNamesRules.rst
ligiabernardet: Underneath "Medium" it says "Suffixes" and, underneath that, there are a number of standard names. Looks ok to me as is.

StandardNamesRules.rst
ligiabernardet: @cacraigucar This has been addressed. Pls take a look.

StandardNamesRules.rst
ligiabernardet: @cacraigucar This has been addressed. Pls take a look.

StandardNamesRules.rst
cacraigucar: Sorry - my reading mistake.  I thought it was a placeholder for things that weren't suffixes or prefixes.

cacraigucar (approved):

StandardNamesRules.rst
gold2718: This is good (discussed at 2021-07-27 framework meeting).

gold2718 (approved):

PR #13, opened 2021-05-07
Updated CCPP Standard Names

bluefinweiwei:
The following two files were changed with the CCPP standard names that are contained under [non-category] section.
	modified:   standard_names.xml
	modified:   Metadata-standard-names.md
fixes #12

PR #16, opened 2021-07-30
Clarification wrt mixing ratios

ligiabernardet:
In these PR I clarified the use of (mass and volume) mixing ratios, as well as the representation of soil moisture. These clarifications are meant to address comments submitted during the review of  https://github.com/ESCOMP/CCPPStandardNames/pull/13.

### DIFF
--- a/StandardNamesRules.rst
+++ b/StandardNamesRules.rst
@@ -56,14 +56,23 @@ CCPP Standard Name Rules
    representing quantities at the interface between grid cells vertically,
    use at_interface.
 
-#. By default, *mixing_ratio* refers to mass mixing ratios. Volume mixing ratios should be
-   qualified as *volume_mixing_ratio*. Mass mixing ratios should contain information regarding 
+#. By default, *mixing_ratio* refers to mass mixing ratios. The long name should
+   explicitly specify that it refers to the *mass* mixing ratio. 
+   Mass mixing ratios should contain information regarding 
    with respect to what quantity they are defined, and options are *wrt_dry_air*, *wrt_moist_air*, 
    or *wrt_total_mass*, where *moist_air* refers to dry air plus vapor and *total_mass* refers
    to dry air plus vapor and hydrometeors. A special case exists in the case of mixing ratio 
    of vapor: the standard name *specific_humidity* should be used instead of 
-   *mixing_ratio_of_vapor_wrt_moist_air*. When referring to soil quantities, 
+   *mixing_ratio_of_vapor_wrt_moist_air*. 
+
+#. Volume mixing ratios should be qualified as *volume_mixing_ratio*.
+
+#. When referring to soil quantities, 
    *volume_fraction* should be used to express the volumetric soil moisture.
+
+#. Number concentration should appear as a prefix, that is, *number_concentration_of*. By default, 
+   number concentrations are specified per unit of volume. When they are specified per
+   unit of mass, they should be written as *mass_number_concentration_of*. 
 
 #. If possible, qualifiers should be limited in order to allow for a wide
    applicability of the variable. In other words, don't qualify with _for ``_xyz``
--- a/StandardNamesRules.rst
+++ b/StandardNamesRules.rst
@@ -56,6 +56,15 @@ CCPP Standard Name Rules
    representing quantities at the interface between grid cells vertically,
    use at_interface.
 
+#. By default, *mixing_ratio* refers to mass mixing ratios. Volume mixing ratios should be
+   qualified as *volume_mixing_ratio*. Mass mixing ratios should contain information regarding 
+   with respect to what quantity they are defined, and options are *wrt_dry_air*, *wrt_moist_air*, 
+   or *wrt_total_mass*, where *moist_air* refers to dry air plus vapor and *total_mass* refers
+   to dry air plus vapor and hydrometeors. A special case exists in the case of mixing ratio 
+   of vapor: the standard name *specific_humidity* should be used instead of 
+   *mixing_ratio_of_vapor_wrt_moist_air*. When referring to soil quantities, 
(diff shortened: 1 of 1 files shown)


ligiabernardet:
@gold2718 @climbfuji @dudhia

climbfuji (approved):
I like this! Thanks!

cacraigucar:
Adding @mattldawson as he pointed out the inconsistency initially.

StandardNamesRules.rst
```diff
@@ -56,6 +56,15 @@ CCPP Standard Name Rules
    representing quantities at the interface between grid cells vertically,
    use at_interface.
 
+#. By default, *mixing_ratio* refers to mass mixing ratios. Volume mixing ratios should be
+   qualified as *volume_mixing_ratio*. Mass mixing ratios should contain information regarding
```
dudhia, line 60: Volume mixing ratio sentence may not be needed as examples of this would be unlikely in microphysics schemes.

ligiabernardet: Volume mixing ratio appear for chemical compositions, and are important in radiation parameterizations. We already have several standard names that start with _volume_mixing_ratio_ (of co2, n2o, ch4, etc.). So, I think this sentence is needed.

StandardNamesRules.rst
```diff
@@ -56,6 +56,15 @@ CCPP Standard Name Rules
...
 
+#. By default, *mixing_ratio* refers to mass mixing ratios. Volume mixing ratios should be
+   qualified as *volume_mixing_ratio*. Mass mixing ratios should contain information regarding 
+   with respect to what quantity they are defined, and options are *wrt_dry_air*, *wrt_moist_air*, 
+   or *wrt_total_mass*, where *moist_air* refers to dry air plus vapor and *total_mass* refers
+   to dry air plus vapor and hydrometeors. A special case exists in the case of mixing ratio 
+   of vapor: the standard name *specific_humidity* should be used instead of 
+   *mixing_ratio_of_vapor_wrt_moist_air*. When referring to soil quantities, 
+   *volume_fraction* should be used to express the volumetric soil moisture.
+
```
dudhia, line 67: Soil moisture should be a separate point from the microphysics paragraph above.

dudhia:
Maybe some separation of microphysics, radiation, and soil paragraphs could
be done.

StandardNamesRules.rst
gold2718: I would still like the term, "mass mixing ratio", to appear in the field's long_name property. Could this be added?

Also, as long as we are at it, should we add a rule for number concentration? The usage in #13 sometimes has the term last (e.g., `prescribed_cloud_droplet_number_concentration`) and sometimes it is first (e.g., `mass_number_concentration_of_aerosol_from_gocart_climatology`). CF seems to always have it first (although it does not have any terms called `mass_number_concentration`).

gold2718 (changes requested):
I have a couple of questions and one suggested change.

StandardNamesRules.rst
ligiabernardet: While these rules are for standard names, we can add a recommendation for using the word "mass" in the long name.

StandardNamesRules.rst
ligiabernardet: OK, we can put the soil aspect in a different bullet

StandardNamesRules.rst
```diff
@@ -478,6 +486,8 @@ standard names.
 +-------------------------------------------+-----------------+
 | volume_fraction                           | 1               |
 +-------------------------------------------+-----------------+
+| volume_mixing_ratio                       | mol mol-1       |
++-------------------------------------------+-----------------+
```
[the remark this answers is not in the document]
ligiabernardet: I asked Mike Iacono of AER, is involved with the RRTMG development, and he recommended mol mol-1. These units also appear in various sites when I do a web search, such as https://www.e-education.psu.edu/meteo300/node/534 or https://www.sciencedirect.com/topics/earth-and-planetary-sciences/mixing-ratio. I wouldn't say these sites are authoritative, but that seems to be how scientists use this quantity.

ligiabernardet:
Regarding @gold2718 's comment about number concentration, we can indeed clarify this in the rules. We likely want _number_concentration_of_X_, to match CF conventions. For number concentration, per unit of volume (m-3) is the default and if the quantity is per unit of mass (kg -1), a qualifier must be used: _mass_number_concentration_.

StandardNamesRules.rst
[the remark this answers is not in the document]
gold2718: @mattldawson, @AndrewJConley, are these units okay with you? Is `volume_mixing_ratio` better than `molar_mixing_ratio`?

I guess that the volume mixing ratio has the same value as the molar mixing ratio as long as the ideal gas equation holds.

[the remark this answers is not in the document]
mattldawson: In my experience, `volume_mixing_ratio` with units of `mol mol-1` is most common (with the assumption of an ideal gas), but I will ask the chemists about this.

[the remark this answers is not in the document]
mattldawson: Louisa agrees and votes for `volume_mixing_ratio` in units of `mol mol-1`

ligiabernardet:
All comments have been addressed in the last update. Please take a look.

StandardNamesRules.rst
gold2718: I am okay with this change, @dudhia, please "Resolve" this conversation if you agree.

gold2718 (approved):
Looks good now, thanks!

StandardNamesRules.rst
dudhia: Looks fine now.

PR #20, opened 2021-08-20
Supercedes #13 + updates to flags & mixing ratios

grantfirl:
This is basically PR #13 + 1 commit that fixes "flag" standard names and various mixing ratios. Commit d4d2c77 is the only difference between this and PR #13, but this PR is started since @bluefinweiwei is on PTO and the PR branch from #13 is in her fork.

Please only review commit d4d2c77 if you want this merged today. Otherwise, I can submit a PR with this commit into Weiwei's branch and we can continue with PR#13 next week. That commit contains the following previously-discussed changes:

 - change "constituent_mixing_ratio" to "potentially_advected_quantities" for lack of better alternatives
 - change many "flag_for..." standard names to "do...", "is...", "disable..." as appropriate
 - change many "flag" units (for integer/reals) to index or none as appropriate
 - change "mixing_ratio" for UFS variables to "mixing_ratio_wrt_moist_air"
 - use mol mol-1 for volume_mixing_ratio names
 - change ...number_concentration to number_concentration_of... for consistency with rules


grantfirl:
@ligiabernardet

grantfirl:
> Why units of none instead of index like the other control variables?

This was a judgement call. I found it odd to give something with a floating point data type an "index" unit. I can change this if desired.

climbfuji:
if it's a number you can always default to unit '1'

cacraigucar:
> > Why units of none instead of index like the other control variables?
> 
> This was a judgement call. I found it odd to give something with a floating point data type an "index" unit. I can change this if desired.

I missed that detail.  I agree that index should be reserved for integer values.

grantfirl:
Closed since Weiwei merged the changed commit into her PR branch... Discussion/review should return to PR #13 now, I guess.

PR #21, opened 2021-08-23
Update standard name rules with rules for units

climbfuji:
Add a section about rules for variables to the standard name rules. In particular, provide guidance for dimensionless variables. 

Disclaimer. I created this PR on GitHub without actually rendering the rst file.

See https://github.com/ESCOMP/CCPPStandardNames/issues/19 and https://github.com/NCAR/ccpp-framework/issues/397 for discussions related to this PR.

### DIFF
--- a/StandardNamesRules.rst
+++ b/StandardNamesRules.rst
@@ -553,7 +553,8 @@ Units
 | flag                   | logicals/booleans that can be either true or false                                  |
 | index                  | integers that can be an index in an array                                           |
 | kg kg-1                | mass mixing ratios                                                                  |
-| mol mol-1              | volume mixing ratios                                                                |
+| m3 m-3                 | volumetric mixing ratios (e.g. for soil moisture)                                   |
+| mol mol-1              | molar mixing ratios (also volumetric mixing ratio for gases)                        |
 | none                   | strings and character arrays                                                        |
 | 1                      | any number (integer, real, complex), e.g. fractions, scaling factors, ...           |
 +------------------------+-------------------------------------------------------------------------------------+
--- a/StandardNamesRules.rst
+++ b/StandardNamesRules.rst
@@ -538,23 +538,30 @@ Units
 =====
 
 #. For variables with an existing match in the `Climate and Forecast (CF) metadata
-   conventions <https://cfconventions.org/standard-names.html>`, the units should
+   conventions <https://cfconventions.org/standard-names.html>`_, the units should
    be identical to the canonical units listed there
 
 #. For variables without an existing match in the CF conventions, the units should
-   follow the International System of Units (SI/metric system), see https://www.nist.gov/pml/weights-and-measures/metric-si/si-units
-   
+   follow the `International System of Units (SI/metric system) <https://www.nist.gov/pml/weights-and-measures/metric-si/si-units>`_
+
 #. For dimensionless variables, the following units can be used:
 
 +------------------------+-------------------------------------------------------------------------------------+
 | **Unit**               |  **Use case**                                                                       |
 +========================+=====================================================================================+
 | count                  | integers that describe the dimension/length of an array                             |
++------------------------+-------------------------------------------------------------------------------------+
 | flag                   | logicals/booleans that can be either true or false                                  |
++------------------------+-------------------------------------------------------------------------------------+
 | index                  | integers that can be an index in an array                                           |
++------------------------+-------------------------------------------------------------------------------------+
 | kg kg-1                | mass mixing ratios                                                                  |
-| m3 m-3                 | volume fraction (e.g. for soil moisture)                                   |
++------------------------+-------------------------------------------------------------------------------------+
+| m3 m-3                 | volume fraction (e.g. for soil moisture)                                            |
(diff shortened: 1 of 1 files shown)


climbfuji:
@ligiabernardet @dudhia I just invited you as collaborators for CCPPStandardNames so that I can add you as reviewers. In the meantime, please see this PR for a proposed extension of the standard name rules regarding units.

cacraigucar (approved):
This is very clear and a welcome addition to the StandardNames

StandardNamesRules.rst
```diff
@@ -532,3 +533,27 @@ Acronyms, Abbreviations, and Aliases
...
+   
+#. For dimensionless variables, the following units can be used:
+
++------------------------+-------------------------------------------------------------------------------------+
+| **Unit**               |  **Use case**                                                                       |
++========================+=====================================================================================+
+| count                  | integers that describe the dimension/length of an array                             |
+| flag                   | logicals/booleans that can be either true or false                                  |
+| index                  | integers that can be an index in an array                                           |
+| kg kg-1                | mass mixing ratios                                                                  |
```
dudhia, line 555: I will be picky here and say mol mol-1 is a molar mixing ratio (which is also a volumetric mixing ratio for gases). However m3 m-3 is a true volumetric mixing ratio unit, so choose to keep the name or the unit.
Soil moisture is often a volumetric mixing ratio and would be m3 m-3

climbfuji: Good point, Jimy, let me make this change ...

dudhia: I will also note that vmr and mmr are often used in the code names but mmr is mass mixing ratio which may be why they use vmr for molar mixing ratio.

climbfuji: @dudhia done, please check

StandardNamesRules.rst
```diff
@@ -532,3 +533,28 @@ Acronyms, Abbreviations, and Aliases
...
+| m m-3                  | volumetric mixing ratios (e.g. for soil moisture)                                   |
```
dudhia, line 556: should be m3 m-3 - seems OK otherwise

climbfuji: Apologies, stupid mistake. Should be fixed now. Please approve if satisfied. Thanks!

dudhia (approved):
this looks fine now

dudhia:
With this new index definition, will the previous real 'index' types be
corrected in the standard names file?
Will there be a new category unit that should be added here?

climbfuji:
> With this new index definition, will the previous real 'index' types be corrected in the standard names file? Will there be a new category unit that should be added here?
> […](#)

The way I understood our previous conversations was that we accepted there are issues with these reals, and we (... well, me) are going to fix this as soon as possible. In the meantime, we chose to "ignore".

dudhia:
OK, need to remember later.

gold2718 (approved):

StandardNamesRules.rst
```diff
@@ -532,3 +533,28 @@ Acronyms, Abbreviations, and Aliases
...
+
+#. For variables without an existing match in the CF conventions, the units should
+   follow the International System of Units (SI/metric system), see https://www.nist.gov/pml/weights-and-measures/metric-si/si-units
+   
+#. For dimensionless variables, the following units can be used:
+
++------------------------+-------------------------------------------------------------------------------------+
+| **Unit**               |  **Use case**                                                                       |
++========================+=====================================================================================+
+| count                  | integers that describe the dimension/length of an array                             |
```
ligiabernardet, line 552: The table is not rendered correctly. Please use other tables as a template, and note that there must be a dashed line between each lines of the table.

StandardNamesRules.rst
```diff
@@ -532,3 +533,28 @@ Acronyms, Abbreviations, and Aliases
...
+| flag                   | logicals/booleans that can be either true or false                                  |
+| index                  | integers that can be an index in an array                                           |
+| kg kg-1                | mass mixing ratios                                                                  |
+| m3 m-3                 | volumetric mixing ratios (e.g. for soil moisture)                                   |
```
ligiabernardet, line 556: This is at odds with rule number 7 listed above, which says: _When referring to soil quantities, volume_fraction should be used to express the volumetric soil moisture._
```suggestion
| m3 m-3                 | volume fraction (e.g. for soil moisture)                                   |
```

ligiabernardet (changes requested):
Thank you for this effort in clarifying the units, especially the dimensionless ones.

StandardNamesRules.rst
climbfuji: Thanks, should be ok now. I rendered it on my mac using `rst2html.py StandardNamesRules.rst  > StandardNamesRules.html` and the results looked ok.

dudhia:
Good point. Soil moisture is m3 m-3 but not a volumetric mixing ratio,
instead a volumetric fraction.
Mixing ratio makes no sense in this context (sorry I think I confused this
issue above).

ligiabernardet (approved):

grantfirl (approved):

PR #23, opened 2021-10-19
Correct spelling of "janic" to "janjic"

climbfuji:
Three standard names are wrong in the dictionary, "janic" needs to be "janjic" (thanks @dusanjovic-noaa for spotting this). 

The corresponding ccpp-physics, fv3atm and ccpp-scm PRs are:

https://github.com/NCAR/ccpp-physics/pull/766
https://github.com/NOAA-EMC/fv3atm/pull/416
https://github.com/NCAR/ccpp-scm/pull/279

For regression testing with the UFS, see https://github.com/ufs-community/ufs-weather-model/pull/892.

### DIFF
--- a/Metadata-standard-names.md
+++ b/Metadata-standard-names.md
@@ -372,9 +372,9 @@ Standard / required CCPP variables
     * `real(kind=kind_phys)`: units = 1
 * `directory_for_rte_rrtmgp_source_code`: Directory for rte rrtmgp source code
     * `character(kind=len=128)`: units = none
-* `do_mellor_yamada_janic_pbl_scheme`: Do mellor yamada janic pbl scheme
+* `do_mellor_yamada_janjic_pbl_scheme`: Do mellor yamada janjic pbl scheme
     * `logical(kind=)`: units = flag
-* `do_mellor_yamada_janic_surface_layer_scheme`: Do mellor yamada janic surface layer scheme
+* `do_mellor_yamada_janjic_surface_layer_scheme`: Do mellor yamada janjic surface layer scheme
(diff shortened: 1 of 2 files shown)


gold2718 (approved):
These changes are okay but I still do not like the proliferation of host-model-specific standard names in this dictionary.

gold2718:
I think that tagging this repo in coordination with tagging a physics repo would be good practice to document the standard names that were used. Something like:
1. Create new tag for CCPPStandardNames
2. Update phyics repo documentation with new tag name
3. tag physics repo

Does that sound right? I think this captures what you are proposing, we just need to formalize it somewhere.

climbfuji:
@gold2718 At the last CCPP framework meeting we discussed that we could merge this PR now, create a new tag and then add that version information to the ccpp-physics main branch. We all agreed that versioning the physics repo every time a new tag is created for CCPP standard names is probably not the best way to go; instead, we were thinking of adding a blob to the top-level README.md that says that this version of ccpp-physics utilizes version x.y.z of the CCPPStandardNames repository. ccpp-physics changes very often, much more often than the standard names repo, and there wouldn't be a way to tell for a certain hash of ccpp-physics which version of the ccpp standard names repo it corresponds to. Does this make sense?

gold2718:
> Does this make sense?

Yes, I thought that is what we agreed to. It allows the two repos to be updated asynchronously.

climbfuji:
> > Does this make sense?
> 
> Yes, I thought that is what we agreed to. It allows the two repos to be updated asynchronously.

Ok, thanks. I will merge this PR.

PR #28, opened 2023-01-06
Add JEDI UFO CF-compliant names

svahl991:
The [JCSDA JEDI software project](https://www.jcsda.org/jcsda-project-jedi) has decided to adopt the CCPP naming standard to use for model variables in its generic code. This PR adds some variable names to CCPP that are already CF-standards and are in use in JEDI.

This is my first attempt at a PR for this repository, so apologies if I haven't done it quite right. I'm happy to modify.

### DIFF
--- a/StandardNamesRules.rst
+++ b/StandardNamesRules.rst
@@ -68,6 +68,10 @@ CCPP Standard Name Rules
 
 #. Volume mixing ratios should be qualified as *volume_mixing_ratio*.
 
+#. By default, *mole_fraction_of_X_in_Y* refers to the total amount of *Y*. So, for example,
+   *mole_fraction_of_ozone_in_air* refers to the total amount of air. When this is not the case,
+   a qualifier should be used to denote this. *e.g.*, *mole_fraction_of_ozone_in_dry_air*.
+
 #. When referring to soil quantities, 
    *volume_fraction* should be used to express the volumetric soil moisture.
 
--- a/StandardNamesRules.rst
+++ b/StandardNamesRules.rst
@@ -69,8 +69,9 @@ CCPP Standard Name Rules
 #. Volume mixing ratios should be qualified as *volume_mixing_ratio*.
 
 #. By default, *mole_fraction_of_X_in_Y* refers to the total amount of *Y*. So, for example,
-   *mole_fraction_of_ozone_in_air* refers to the total amount of air. When this is not the case,
-   a qualifier should be used to denote this. *e.g.*, *mole_fraction_of_ozone_in_dry_air*.
+   *mole_fraction_of_ozone_in_air* refers to the total amount of (moist) air. (In the case of air,
+   the default meaning is moist air, as described in the *mixing ratio* rule.) When this is not
+   the case, a qualifier should be used to denote this. *e.g.*, *mole_fraction_of_ozone_in_dry_air*. 
 
 #. When referring to soil quantities, 
    *volume_fraction* should be used to express the volumetric soil moisture.
--- a/standard_names.xml
+++ b/standard_names.xml
@@ -365,6 +365,9 @@
     <standard_name name="mole_fraction_of_ozone_in_air">
       <type kind="kind_phys" units="mol mol-1">real</type>
     </standard_name>
+    <standard_name name="mole_fraction_of_carbon_dioxide_in_air">
+      <type kind="kind_phys" units="mol mol-1">real</type>
+    </standard_name>
     <standard_name name="cloud_liquid_water_mixing_ratio_wrt_moist_air">
                    long_name="Ratio of the mass of liquid water to the mass of moist air">
       <type kind="kind_phys" units="kg kg-1">real</type>
--- a/standard_names.xml
+++ b/standard_names.xml
@@ -171,6 +171,14 @@
                    long_name="Horizontal wind in a direction perdendicular to x_wind">
       <type kind="kind_phys" units="m s-1">real</type>
     </standard_name>
+    <standard_name name="eastward_wind"
+                   long_name="Wind vector component, positive when directed eastward">
+      <type kind="kind_phys" units="m s-1">real</type>
+    </standard_name>
+    <standard_name name="northward_wind"
+                   long_name="Wind vector component, positive when directed northward">
+      <type kind="kind_phys" units="m s-1">real</type>
+    </standard_name>
     <standard_name name="dry_static_energy"
                    long_name="Dry static energy Content of Atmosphere Layer">
       <type kind="kind_phys" units="J kg-1">real</type>
--- a/standard_names.xml
+++ b/standard_names.xml
@@ -362,6 +362,9 @@
     <standard_name name="mole_fraction_of_water_vapor">
       <type kind="kind_phys" units="mol mol-1">real</type>
     </standard_name>
+    <standard_name name="mole_fraction_of_ozone_in_air">
+      <type kind="kind_phys" units="mol mol-1">real</type>
+    </standard_name>
     <standard_name name="cloud_liquid_water_mixing_ratio_wrt_moist_air">
                    long_name="Ratio of the mass of liquid water to the mass of moist air">
       <type kind="kind_phys" units="kg kg-1">real</type>
--- a/standard_names.xml
+++ b/standard_names.xml
@@ -323,6 +323,10 @@
                    long_name="air potential temperature on previous timestep">
       <type kind="kind_phys" units="K">real</type>
     </standard_name>
+    <standard_name name="virtual_temperature"
+                   long_name="virtual temperature">
+      <type kind="kind_phys" units="K">real</type>
+    </standard_name>
     <standard_name name="pressure_dependent_gas_constant_of_dry_air">
       <type kind="kind_phys" units="J kg-1 K-1">real</type>
     </standard_name>
--- a/standard_names.xml
+++ b/standard_names.xml
@@ -352,6 +352,9 @@
     <standard_name name="cloud_area_fraction_in_atmosphere_layer">
       <type kind="kind_phys" units="1">real</type>
     </standard_name>
+    <standard_name name="relative_humidity">
+      <type kind="kind_phys" units="1">real</type>
+    </standard_name>
   </section>
   <section name="diagnostics">
     <standard_name name="total_precipitation_rate_at_surface">
--- a/standard_names.xml
+++ b/standard_names.xml
@@ -383,6 +383,10 @@
     <standard_name name="mole_fraction_of_carbon_dioxide_in_air">
       <type kind="kind_phys" units="mol mol-1">real</type>
     </standard_name>
+    <standard_name name="humidity_mixing_ratio">
+                   long_name="Ratio of the mass of water vapor to the mass of dry air">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
     <standard_name name="cloud_liquid_water_mixing_ratio_wrt_moist_air">
                    long_name="Ratio of the mass of liquid water to the mass of moist air">
       <type kind="kind_phys" units="kg kg-1">real</type>
--- a/standard_names.xml
+++ b/standard_names.xml
@@ -383,7 +383,7 @@
     <standard_name name="mole_fraction_of_carbon_dioxide_in_air">
       <type kind="kind_phys" units="mol mol-1">real</type>
     </standard_name>
-    <standard_name name="humidity_mixing_ratio">
+    <standard_name name="water_vapor_mixing_ratio_wrt_dry_air">
                    long_name="Ratio of the mass of water vapor to the mass of dry air">
       <type kind="kind_phys" units="kg kg-1">real</type>
     </standard_name>
--- a/standard_names.xml
+++ b/standard_names.xml
@@ -349,6 +349,9 @@
     <standard_name name="mass_content_of_cloud_liquid_water_in_atmosphere_layer">
       <type kind="kind_phys" units="kg m-2">real</type>
     </standard_name>
+    <standard_name name="cloud_area_fraction_in_atmosphere_layer">
+      <type kind="kind_phys" units="1">real</type>
+    </standard_name>
   </section>
   <section name="diagnostics">
     <standard_name name="total_precipitation_rate_at_surface">
--- a/standard_names.xml
+++ b/standard_names.xml
@@ -224,9 +224,6 @@
     <standard_name name="geopotential_height">
       <type kind="kind_phys" units="m">real</type>
     </standard_name>
-    <standard_name name="height">
-      <type kind="kind_phys" units="m">real</type>
-    </standard_name>
     <standard_name name="potentially_advected_quantities">
       <type kind="kind_phys" units="various">real</type>
     </standard_name>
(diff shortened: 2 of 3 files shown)


standard_names.xml
```diff
@@ -350,6 +377,16 @@
     <standard_name name="mole_fraction_of_water_vapor">
       <type kind="kind_phys" units="mol mol-1">real</type>
     </standard_name>
+    <standard_name name="mole_fraction_of_ozone_in_air">
+      <type kind="kind_phys" units="mol mol-1">real</type>
+    </standard_name>
+    <standard_name name="mole_fraction_of_carbon_dioxide_in_air">
+      <type kind="kind_phys" units="mol mol-1">real</type>
+    </standard_name>
```
gold2718, line 385: Should (does) this specify `dry`, `moist`, or `wet` air?
I realize that these are CF standard names, however, I do not understand what the specific meaning of `_in_air` means in CF. Can anyone enlighten me here?

standard_names.xml
```diff
@@ -350,6 +377,16 @@
...
+    <standard_name name="humidity_mixing_ratio">
```
gold2718, line 386: This standard name is not following the CCPP rules which would be.
```
    <standard_name name="water_vapor_mixing_ratio_wrt_dry_air">
```
On the other hand, `humidity_mixing_ratio` _is_ a CF standard name. So by rule 1, this is the correct name. Fun!

gold2718 (commented):
I am only commenting on the xml file since the md file is generated.
These additions look like they are following the rules, however, in one case, I'm not sure what the more specific CCPP version would be.
One solution is to add the full and specific definition as a `long_name` attribute.
On the other hand, if the CF name is _not_ specific, I think we should switch to using the CCPP standard.
Thoughts?

standard_names.xml
```diff
@@ -171,6 +171,14 @@
...
       <type kind="kind_phys" units="m s-1">real</type>
     </standard_name>
+    <standard_name name="eastward_wind"
+                   long_name="Wind vector component, positive when directed eastward">
+      <type kind="kind_phys" units="m s-1">real</type>
+    </standard_name>
+    <standard_name name="northward_wind"
+                   long_name="Wind vector component, positive when directed northward">
+      <type kind="kind_phys" units="m s-1">real</type>
+    </standard_name>
```
cacraigucar, line 181: NOAA and NCAR had extensive discussions about how to label winds in a generic way a number of years ago.  With various grids where the wind components don't follow a east/west lines, we had decided that a generic x/y coordinate label was preferable.  @grantfirl - What are your thoughts?  @svahl991 - Would you folks be willing to adopt this more generic name?

standard_names.xml
```diff
@@ -216,6 +224,9 @@
     <standard_name name="geopotential_height">
       <type kind="kind_phys" units="m">real</type>
     </standard_name>
+    <standard_name name="height">
```
cacraigucar, line 227: We already have height_above_main_sea_level Would this variable be identical?

standard_names.xml
```diff
@@ -331,6 +346,18 @@
...
+    </standard_name>
+    <standard_name name="mass_content_of_cloud_liquid_water_in_atmosphere_layer">
+      <type kind="kind_phys" units="kg m-2">real</type>
+    </standard_name>
+    <standard_name name="cloud_area_fraction_in_atmosphere_layer">
+      <type kind="kind_phys" units="1">real</type>
+    </standard_name>
+    <standard_name name="relative_humidity">
+      <type kind="kind_phys" units="1">real</type>
+    </standard_name>
```
cacraigucar, line 360: I was surprised that we didn't already have `relative_humidity`, but I don't see something which I think would be it currently in our list.  @grantfirl - do you agree?

standard_names.xml
svahl991: The question of whether JEDI is willing to adopt x/y instead of east/north for winds is probably above my pay grade. I can imagine a situation where a particular JEDI algorithm would work with east/north winds but not with all more generic x/y winds, and so JEDI would want to be able to distinguish between the two, but I don't know if that is a real situation or not. I can ask others with more expertise about this.

I will also note that such a change does come with a cost to more than just the core JEDI codebase. JEDI is used by many different atmospheric models, not just NOAA and NCAR models, all of which will need to change or translate existing east/north names in their code in order to communicate with JEDI if JEDI does not accept east/north names. There are cases where JEDI is using non-CCPP-compliant variable names where we already know we will need to incur this difficulty, but I was hoping it could be avoided for names that are already CF-compliant. I thought that a CF-compliant name was also a CCPP-compliant name, based on Rule number 1, but based on the discussions I'm seeing in this PR, it appears that might not be true. Perhaps rule number 1 could be clarified?

climbfuji: I have the same concern that @svahl991 has w.r.t. that some algorithm may be relying on eastward and northward, and not generic x/y directions. I know that MPAS for example computes east/north winds from its generic wind directions (which are orthogonal to the cell faces) to pass them to the physics, but is JEDI using those east/north winds or the "native" wind directions from the MPAS mesh?

If JEDI does require true east and north, and not generic x and y, then that's what it should be using as standard names. If it doesn't need true east and north, but arbitrary orthogonal x and y, then it should be using x and y. My 2c.

cacraigucar: I second @climbfuji's take on east/west versus x/y.  If a package needs east/west then it should use those specific names.  But if it happens to be east/west but would work for any orthogonal directions, for flexibility sake it should use the x/y names.

Also, the standard names are only in the CCPP metadata.  If internally code wants to continue using the names with east/west, they can do that and link them to the x/y CCPP names in the metadata.   This means that I don't think code rewrites need to be done to change internal variable names.

grantfirl: I agree. If the variables truly represent east/west, then those names should be kept. CF doesn't specify, but I'm assuming that east/west is in reference to geographic direction as opposed to magnetic or some other definition. Perhaps zonal and meridional would work better to avoid any confusion, but those aren't actually CF names. Weird.

standard_names.xml
grantfirl: Ya, this is exhibit A why we've been extending CF.

standard_names.xml
grantfirl: Yes. UFS uses relative_humidity, but it is in the interstitial DDT which we yet to parse for standard name rules and added to this repo. It should absolutely be added.

standard_names.xml
cacraigucar: After discussion with @nusbaume here on the NCAR side, we agree with @gold2718 that the standard name should be `water_vapor_mixing_ratio_wrt_dry_air` and not the `humidity_mixing_ratio` which is originally proposed.

standard_names.xml
svahl991: We discussed this more with a group here at the JCSDA and we feel we do need to keep the east/west names as the x/y names are too general for all our needs.

climbfuji: To add a bit more information, there is an identity operator in JEDI for obs and those obs report geographical directions, therefore those specific names are needed.

standard_names.xml
BenjaminTJohnson: Shouldn't these also specify whether it's a "volume mixing ratio" or a "mass mixing ratio"?

svahl991: > Shouldn't these also specify whether it's a "volume mixing ratio" or a "mass mixing ratio"?

By rule number 5 here: https://github.com/ESCOMP/CCPPStandardNames/blob/main/StandardNamesRules.rst, I think the default is to refer to mass mixing ratio unless otherwise specified.

climbfuji: Correct

standard_names.xml
gthompsnJCSDA: If there is already a CCPP height variable with the name `height_above_mean_sea_level` then we should not confuse people with an added variable named only `height`.  Notice the spelling of `main` in the comment from @cacraigucar appears to be a mistake and should be as I typed it with `mean_sea_level` in the name.

cacraigucar: Yup - a typo - I've corrected

svahl991: OK. I will remove `height` from this PR.

standard_names.xml
cacraigucar: It appears that this use of east/west is appropropriate, so I will resolved this conversation

standard_names.xml
cacraigucar: Once the removal of `height` is pushed back to this PR, I will resolve this conversation.  This appears to be the last outstanding comment I have before I can give my approval.

standard_names.xml
cacraigucar: @fvitt and/or @mattldawson - From a chemistry standpoint do either of you have any insights on how folks in your community explicitily describe `mole_fractions`?

svahl991: Thanks to @huishao-r, who found this description for me:
https://vocab.nerc.ac.uk/collection/P07/current/CFV7N47/
Is this helpful? I don't know if it explicitly answers the dry/moist/wet question. My guess is that it is assumed to just be in the atmosphere. Does dry/moist/wet need to be specified in all contexts?

BenjaminTJohnson: I'm not sure if this is helpful or not, but we never use that unit inside of CRTM.

gthompsnJCSDA: I will throw in my 2 cents here.  We live in a moist world.  The air around us is filled with molecules of water vapor alongside CO2, O2, N2, O3, etc.  For those models that use a "dry_air" reference point, then I say make those models add "dry_air" to their wording otherwise "in_air" can be interpreted as being with reference to moisture included since that is the physical world we live in.  The word "wet" being used is so highly confusing.

Molecules of various constituents simply coexist.  We could have moles of water vapor per mole of air, moles of CO per mole of air, moles of CO2 per mole of air, etc.  Taking one type of molecule type out (i.e., water vapor) makes zero sense because it's like defining another constituent after taking away the N2 molecules or something.  There's no end to the possibilities.  So the definition - **again in my opinion** - should be consistent with the physical world of air is a mixture of gases so we shouldn't define a variable with regard to dry (non-H2o) air if we are not eliminating the other constituents.

I think I'm showing my frustration towards model concepts that don't relate well to real world. If a modeler cannot output the data into variables of a physical world by simply including H2O molecules in air, then why not?  They don't go around excluding the other constituents so poor water vapor is always picked on :)

dudhia: From a physics perspective this would be relative to total air, the way ppm by volume is, and it would be directly proportional to ppmv.

svahl991: Thanks to everyone for chiming in on this discussion today! After these clarifications, is there still a question we're trying to answer here? If so, can it be stated? Of course the question I'm most interested in is: will this PR be approved with these two `mole_fraction_*` variables named as they are?

dudhia: Based on my comment, this name is sufficient.

standard_names.xml
```diff
@@ -383,15 +380,15 @@
     <standard_name name="mole_fraction_of_carbon_dioxide_in_air">
       <type kind="kind_phys" units="mol mol-1">real</type>
     </standard_name>
-    <standard_name name="water_vapor_mixing_ratio_wrt_dry_air">
+    <standard_name name="water_vapor_mixing_ratio_wrt_dry_air"
```
cacraigucar, line 383: Oops - you removed the `>` at the end of this variable, along with the next two as well.

cacraigucar (changes requested):

standard_names.xml
svahl991: Yes, I did that intentionally because the `long_name` on the next line is part of the same tag. I think this is the correct syntax.

When I first made the PR, I manually edited the `Metadata-standard-names.md` file, not realizing it should be generated with the Python script, which is why I didn't notice the XML formatting error in this file. @climbfuji is going to generate the `.md` file for me this time, since he has easier access to a python environment where the script will run. So this PR is currently not ready to go since the `.md` file isn't updated yet. He may also find other XML bugs when he runs the script. So we need a little more time.

climbfuji: Should be done now (c1b481d)

standard_names.xml
gold2718: > Does dry/moist/wet need to be specified in all contexts?

While I sympathize with the arguments put forward by @gthompsnJCSDA, physics schemes need to know what physical quantity it is dealing with and a principle of the CCPP has been to provide that clarity.
I am fine with `mole_fraction_*` meaning with respect to total air (as opposed to 'dry air' or 'moist air'), however, if we decide this, can we please add this as a rule in StandardNamesRules.rst? That way, if anyone comes in with moles of X per moles of moist / dry air, they will have to use standard names that reflect this. I would like that addition to be part of this PR.

svahl991:
Apologies if I accidentally removed @ligiabernardet as a reviewer on this PR. I'm not sure how that happened. I don't appear to have the ability to re-add.

climbfuji:
I added her back in. I'll review later today.

gold2718 (approved):
I think this looks great, thanks for all the work.

StandardNamesRules.rst
```diff
@@ -68,6 +68,10 @@ CCPP Standard Name Rules
 
 #. Volume mixing ratios should be qualified as *volume_mixing_ratio*.
 
+#. By default, *mole_fraction_of_X_in_Y* refers to the total amount of *Y*. So, for example,
```
climbfuji, line 71: I am sorry that I need to ask one more time. It reads like total amount of Y in the case of air means total air mass, i.e  w.r.t. dry air + water vapor + all hydrometeors, as opposed to w.r.t. dry air + water vapor only. I just want to make sure that this is the intent.

gthompsnJCSDA: I do not believe the solid particles should be taken into account in this circumstance.  They would seem quite unusual.

climbfuji: > I do not believe the solid particles should be taken into account in this circumstance. They would seem quite unusual.

In this case, the rule should state that? i.e. something like "if Y is air, then the default is "wrt_moist_air" as described in rule 5?

BenjaminTJohnson: I don't know what the general consensus is, but generally mixing ratios are most useful / interesting when dealing with moist air, with the noted exception of water vapor mixing ratio, which is universally applied to dry air. 

Water vapor mixing ratio with respect to moist air is defined as "specific humidity", and they differ by as much as a few percent in the Earth's atmosphere.

BenjaminTJohnson (approved):
Looks like some long names are missing, or inconsistent in scope / style with respect to other long names, but I suppose this is something that will be iteratively added.

StandardNamesRules.rst
svahl991: I have just pushed an update to the wording of the rule to be more clear. See [af15329](https://github.com/ESCOMP/CCPPStandardNames/pull/28/commits/af1532942e96ac01f0df5a416e68cabb38fd39ef).

dudhia:
The distinction is needed because some models refer all species to dry air.

climbfuji (approved):

climbfuji:
I re-requested reviews from reviewers who had comments earlier.

cacraigucar (approved):

climbfuji:
We have four approvals, therefore I am going to merge this and not let it drag on for longer.

PR #29, opened 2023-03-05
Updates to geopotential height and reference pressure

nusbaume:
As discussed in the February 27th, 2023 framework meeting, scientists in NCAR CGD requested that some of the standard names related to geopotential height and reference pressure be modified to either be more explicit or to better match already existing standards.  This PR implements those requests.

Also, I wasn't sure who to request a review from, so I selected semi-randomly based off of previous PRs.  If I missed anyone, or selected someone who doesn't need to review, then please let me know.  Thanks!

### DIFF
--- a/Metadata-standard-names.md
+++ b/Metadata-standard-names.md
@@ -204,7 +204,7 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
 * `reference_pressure`: A constant reference pressure value of 1000 hPa
     * `real(kind=kind_phys)`: units = Pa
 * `reference_pressure_of_atmosphere_layer`: The reference pressure for a given atmosphere layer
-    * `real(kind=kind_phys)`: units = None
+    * `real(kind=kind_phys)`: units = Pa
 * `reference_air_pressure_normalized_by_surface_air_pressure`: reference pressure normalized by surface pressure
     * `real(kind=kind_phys)`: units = 1
 * `dimensionless_exner_function`: exner function
--- a/Metadata-standard-names.md
+++ b/Metadata-standard-names.md
@@ -201,7 +201,7 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
     * `real(kind=kind_phys)`: units = W m-2
 * `US_standard_air_pressure_at_sea_level`: US Standard Atmospheric pressure at sea level
     * `real(kind=kind_phys)`: units = Pa
-* `reference_pressure`: A constant reference pressure value of 1000 hPa
+* `reference_pressure`: reference pressure used in definition of potential temperature, Exner function, etc.
     * `real(kind=kind_phys)`: units = Pa
 * `reference_pressure_in_atmosphere_layer`: reference pressure in atmosphere layer
     * `real(kind=kind_phys)`: units = Pa
--- a/Metadata-standard-names.md
+++ b/Metadata-standard-names.md
@@ -203,7 +203,7 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
     * `real(kind=kind_phys)`: units = Pa
 * `reference_pressure`: A constant reference pressure value of 1000 hPa
     * `real(kind=kind_phys)`: units = Pa
-* `reference_pressure_of_atmosphere_layer`: The reference pressure for a given atmosphere layer
+* `reference_pressure_in_atmosphere_layer`: The reference pressure for a given atmosphere layer
     * `real(kind=kind_phys)`: units = Pa
 * `reference_air_pressure_normalized_by_surface_air_pressure`: reference pressure normalized by surface pressure
     * `real(kind=kind_phys)`: units = 1
--- a/Metadata-standard-names.md
+++ b/Metadata-standard-names.md
@@ -203,7 +203,7 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
     * `real(kind=kind_phys)`: units = Pa
 * `reference_pressure`: A constant reference pressure value of 1000 hPa
     * `real(kind=kind_phys)`: units = Pa
-* `reference_pressure_in_atmosphere_layer`: The reference pressure for a given atmosphere layer
+* `reference_pressure_in_atmosphere_layer`: reference pressure in atmosphere layer
     * `real(kind=kind_phys)`: units = Pa
 * `reference_air_pressure_normalized_by_surface_air_pressure`: reference pressure normalized by surface pressure
     * `real(kind=kind_phys)`: units = 1
--- a/Metadata-standard-names.md
+++ b/Metadata-standard-names.md
@@ -147,7 +147,9 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
     * `real(kind=kind_phys)`: units = 1
 * `reciprocal_of_dimensionless_exner_function_wrt_surface_air_pressure`: inverse exner function w.r.t. surface pressure, (ps/p)^(R/cp)
     * `real(kind=kind_phys)`: units = 1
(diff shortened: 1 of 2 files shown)


nusbaume:
Sorry, missed one commit.  NOW it should be ready for review.  Thanks again!

dudhia:
Is reference pressure just a scalar. I assume so.
I am curious how reference pressure in a layer is used. Also a scalar?

nusbaume:
@dudhia Yes, "reference pressure" is just a scalar, which at least in CAM is assumed to be 1000 hPa (basically the P0 quantity used to calculate potential temperature).

The "reference pressure in a layer" variable is a more CAM-unique quantity.  Basically we have a reference profile of pressure we use in the model for our hybrid sigma-pressure vertical coordinate.  Those pressures don't change in time or horizontal space, but are a function of height, with the scalar "reference pressure" being the starting value at the surface.

Hope that helps clarify the difference, at least a little bit!

dudhia:
@nusbaume OK, thanks, so can we view this as a 1d list of numbers defining coordinate levels?

nusbaume:
@dudhia Yes, at least that is how I understand it.

Metadata-standard-names.md
```diff
@@ -197,11 +199,11 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
...
 * `cumulative_boundary_flux_of_total_water`: Cumulative boundary flux of total water
     * `real(kind=kind_phys)`: units = W m-2
-* `reference_pressure`: reference pressure 
-    * `real(kind=kind_phys)`: units = Pa
 * `US_standard_air_pressure_at_sea_level`: US Standard Atmospheric pressure at sea level
     * `real(kind=kind_phys)`: units = Pa
-* `reference_pressure_at_surface`: reference pressure at surface
+* `reference_pressure`: A constant reference pressure value of 1000 hPa
+    * `real(kind=kind_phys)`: units = Pa
+* `reference_pressure_in_atmosphere_layer`: The reference pressure for a given atmosphere layer
```
cacraigucar, line 206: Would changing the longname help?   Note I am not proposing changing the standard name.   Proposed new longname:   "Reference pressures for atmosphere layers"  This helps to describe that there is more than one and indicates it is not a scalar.

grantfirl: Other standard names in this repo basically just repeat "in atmosphere layer" in the long name. It isn't super important though, because I imagine that when this standard name is used in the wild (in models), it will come with dimensions that make it clear what it is representing. With long names, the principle is basically like the Hippocratic Oath: first, do no harm. So, if the long name makes the standard name less clear, it is better to simply repeat the standard name rather than come up with a synonym that says the same thing, or (gasp), can even be interpreted completely differently.

dudhia: sounds like it might be better

dudhia: OK, the name already relates it to model layers. No change needed.

nusbaume: I am happy to change the long name if it is preferred by the reviewers.  It sounds like folks want either `reference pressures for atmosphere layers` or `reference pressure in atmosphere layer`?

cacraigucar: Either of the proposed changes for the long name are fine with me.  It probably makes sense to go with Grant's suggestion, since there is precedence already.

nusbaume: Changed the long name to `reference pressure in atmosphere layer`.

cacraigucar (approved):

svahl991:
I have no concerns with these changes from a JEDI perspective, but I do have a question: Is there ever a situation where `reference_pressure` might be some other value than 1000 hPa? Or, at least as far as this naming standard is concerned, `reference_pressure` is always that single, fixed value?

gold2718:
> I have no concerns with these changes from a JEDI perspective, but I do have a question: Is there ever a situation where `reference_pressure` might be some other value than 1000 hPa? Or, at least as far as this naming standard is concerned, `reference_pressure` is always that single, fixed value?

I imagine that folks studying other planets (e.g., Mars) would use a (drastically) different value.

svahl991:
>  imagine that folks studying other planets (e.g., Mars) would use a (drastically) different value.

So in that situation, would a different name for the quantity be used? (_e.g._ `reference_pressure_for_mars`) Or is `reference_pressure` not actually a constant value, but a variable one? I'm asking because I deal with generic code that uses this `reference_pressure` value, and the formulas in the code would theoretically work for any reference pressure, so I was treating the value as a variable, not a constant. But then I don't know if I should use the CCPP name `reference_pressure` to refer to the quantity that is in my code, since that value, at least in the CCPP description, is fixed.

nusbaume:
If I remember correctly (?) the scientists in my section were arguing that it should be exactly 1000 hPa because that way one can be sure that there isn't an artificial model dependence being introduced for certain thermodynamic quantities, such as potential temperature (i.e. every model's potential temperature should be relative to 1000 hPa if they are simulating Earth).  @PeterHjortLauritzen am I expressing your views correctly, or am I wrong here?

svahl991:
So if I have a formula in my code to calculate potential temperature, and that formula is generic enough to work (maybe someday) for a Mars atmosphere model, it seems as if I should use some other name besides `reference_pressure` for the quantity in my code. Is there another name already in the CCPP standard for a "generic", or variable, reference pressure quantity?

I hope I'm not being overly pedantic about this since the code is not yet being used for anything but earth models. But we are trying to be precise with these names, and there is a conceptual difference between the (theoretically) generic "p-naught" value in the formula for potential temperature, and the specific numeric value it is usually assigned for earth models. So I'm just wondering if we have two different names for the two concepts.

gold2718:
CESM/CAM has been used for studies of other planets, including Mars. In those studies, the variable was reused with a different value.

However, I see your point that a CCPP Standard Name is supposed to refer to a well-defined physical quantity and one could view the reference surface pressure used on different planets to be different. I would love some more feedback here.

svahl991:
> In those studies, the variable was reused with a different value.

What drew my attention to this in the first place was that this PR adds the specific numerical value for the quantity to the description of the `reference_pressure` variable/constant. Without that, the name remains more generic.

dudhia:
This argues for keeping the original long name
"reference pressure at surface"

PeterHjortLauritzen:
reference_pressure is an arbitrarily selected atmospheric pressure of 1000mb to which adiabatic processes are referred for definitions of potential temperature, Exner pressure, etc. 

For other planets reference_pressure would be a different value. I guess one could use a different value for Earth's atmosphere but there is agreement among meorologists to use 1000hPa.

I would argue to keep the short name "reference_pressure" but change the long name to

"Reference pressure used in definition of potential temperature, Exner function, etc."

nusbaume:
Hi @svahl991 , would the `reference_pressure` long name being:

`Reference pressure used in definition of potential temperature, Exner function, etc.` 

work better for you on the JEDI-side of things?  If so I'll go ahead and make the change.  Thanks!

svahl991:
> Hi @svahl991 , would the `reference_pressure` long name being:
> 
> `Reference pressure used in definition of potential temperature, Exner function, etc.`
> 
> work better for you on the JEDI-side of things? If so I'll go ahead and make the change. Thanks!

Yes, I personally find that name both more descriptive and less planetary-restrictive for generic software like JEDI. Who wants to go to all the trouble of changing the names in the software when we're making daily weather forecasts for our Mars colony? But I'm a software engineer, not a scientist, so I defer to the scientists for final say on this.

svahl991 (approved):

grantfirl (approved):

climbfuji (approved):

PR #32, opened 2023-03-28
Fix bugs in python table writing and checking scripts

### LINKED ISSUE #30: write_standard_name_table.py doesn't work with python 3.7+
Currently the `write_standard_name_table.py` script will not work with later (3.7+) versions of python due to an incompatibility in the subprocess flags used in the `call_command` function contained within `xml_tools.py`.  This can be resolved by simply updating the `xml_tools.py` file to the latest version used in the CCPP-framework.

### LINKED ISSUE #31: check_xml_unique.py --overwrite doesn't actually remove duplicate variables
The `check_xml_unique.py` function's `--overwrite` flag doesn't actually remove the duplicated variables in the overwritten file.  This is because the code section that tries to remove the duplicate names has a typo in its find path, and thus never successfully finds the elements that need to be removed.  Plus it appears that some of the list dimensioning is incorrect, and will result in duplicate elements getting skipped if they are in the same `section`.

nusbaume:
This PR modifies the xml_tools library to allow the use of python 3.7+ when using `write_standard_name_table.py`.  Also fixed the use of the `--overwrite` flag in `check_xml_unique.py`.

Fixes #30 
Fixes #31 

### Tests run:

Ran tests with `check_xml_unique.py` and `check_xml_unique.py --overwrite` where duplicate variables were added to both the same section and to different sections of `standard_names.xml`, to make sure the script worked as expected.

Also ran `write_standard_name_table.py` on a modified `standard_names.xml` file to make sure the Markdown file was created successfully.  This was done with python 3.9.4, 3.10.8, and 2.7.17 to try and test that the script works with multiple versions, including versions newer than 3.6.

Finally, I again wasn't sure who to add as reviewers, so please feel free to expand or modify the reviewer list if need be.

### DIFF
--- a/check_xml_unique.py
+++ b/check_xml_unique.py
@@ -22,7 +22,7 @@ def parse_command_line(args, description):
                         type=str, help="XML file with standard name library")
     parser.add_argument("--overwrite", action='store_true',
                         help="flag to remove duplicates and overwrite the file")
-    
+
     pargs = parser.parse_args(args)
     return pargs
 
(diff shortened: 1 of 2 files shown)


nusbaume:
@mkavulich I sent you an invite to join this repo in order to add you as a reviewer.  Of course if you would prefer to not review this PR (or generally don't want to review PRs in this repo) then just let me know.  Thanks!

mkavulich (approved):
I haven't tested this myself but the code changes look good to me.

mkavulich (approved):

grantfirl (approved):

cacraigucar (approved):
Tested with a duplicate name and it worked as advertised

PR #34, opened 2023-04-27
Add GitHub Actions CI for pull requests

### LINKED ISSUE #33: Add check_xml_unique Github Action
The `check_xml_unique.py` tool can notify a developer when duplicate standard names exist in the dictionary, which is good to avoid since having duplicate standard names introduces an ambiguity which could then cause problems downstream (for example when the CCPP framework tries to generate interface codes).

However, there is no guarantee that a developer will run this tool before pushing changes to the standard names dictionary, which means a duplicate standard name could sneak in on accident.  To avoid this problem, a Github Action workflow should be created that runs `check_xml_unique.py` on the dictionary file (`standard_names.xml`) whenever a Pull Request is opened or updated in this repo, which then should "fail" whenever a duplicate name is found.

Finally, along with the duplicate name check itself, the `check_xml_unique.py` should also run `validate_xml_file` similar to `write_standard_name_table.py` in order to ensure `xmllint` is always run on the dictionary file as well (to avoid any basic syntax errors).

jfrost-mo:
Fixes #33

Also has a second actions job to check that the markdown rendering of the standard names is up to date. It would be fairly easy to extend this to add a commit to the PR updating it if it isn't, but I haven't added that here as it wasn't really the focus.

### DIFF
--- a/.github/workflows/pull_request_ci.yml
+++ b/.github/workflows/pull_request_ci.yml
@@ -13,6 +13,11 @@ jobs:
         with:
           python-version: "3.x"
 
+      - name: Install xmllint
+        run: |
+            sudo apt-get update
+            sudo apt-get -y install libxml2-utils
+
       - name: Check for duplicate standard names
         run: python3 check_xml_unique.py standard_names.xml
 
@@ -24,6 +29,11 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: "3.x"
+
+      - name: Install xmllint
+        run: |
+            sudo apt-get update
+            sudo apt-get -y install libxml2-utils
 
       - name: Check markdown has been rerendered
         run: |
--- a/.github/workflows/pull_request_ci.yml
+++ b/.github/workflows/pull_request_ci.yml
@@ -1,0 +1,35 @@
+name: Pull request checks
+
+on:
+  pull_request:
+
+jobs:
+  check-unique-standard-names:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.x"
+
+      - name: Check for duplicate standard names
+        run: python3 check_xml_unique.py standard_names.xml
+
+  check-rerendered-markdown:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.x"
+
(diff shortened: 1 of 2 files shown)


.github/workflows/pull_request_ci.yml
```diff
@@ -0,0 +1,35 @@
...
+jobs:
+  check-unique-standard-names:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.x"
+
```
nusbaume, line 15: Sadly `xmllint` doesn't come with the standard runner environment so we'll need to add it here:

```suggestion
 # Install xmllint so tests pass:
      - name: Install xmllint
        run: |
            sudo apt-get update
            sudo apt-get install libxml2-utils
```

.github/workflows/pull_request_ci.yml
```diff
@@ -0,0 +1,35 @@
...
+  check-rerendered-markdown:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.x"
+
```
nusbaume, line 27: We'll have to add the `xmllint` install step here as well:

```suggestion
 # Install xmllint so tests pass:
      - name: Install xmllint
        run: |
            sudo apt-get update
            sudo apt-get install libxml2-utils
```

nusbaume (changes requested):
Thanks so much for opening this PR!  I didn't even realize the Met Office was interested in this project, but we'll happily take any help we can get.

Anyways, I tried testing the Actions workflow and found that one step related to xmllint was missing for each job.  However, once those changes are made and the merge conflict is resolved I think this PR will be good to go, at least on my end (i.e. the workflow tests all behaved as expected in my testing environment).  Of course If you need any assistance with the requested changes just let me know.

.github/workflows/pull_request_ci.yml
jfrost-mo: xmllint was installed on my local environment already, so I forgot about installing it. Thanks.

jfrost-mo:
I've added xmllint installation to the workflow, and rebased onto main to fix the merge conflicts.

nusbaume (approved):
Looks good to me now, thanks again!

mkavulich (approved):

climbfuji (approved):
Thanks very much for adding this capability, @jfrost-mo !

PR #35, opened 2023-04-28
Add new 'precipitation' standard name rule.

nusbaume:
The CGD scientists had some debate about what the word "precipitation" should mean in the context of a standard name.  After discussing it in a framework meeting, the following rule proposed in this PR seemed to be a good option.  Please let me know if you think otherwise, or would like any changes.  Thanks!

### DIFF
--- a/StandardNamesRules.rst
+++ b/StandardNamesRules.rst
@@ -58,27 +58,33 @@ CCPP Standard Name Rules
    use at_interface.
 
 #. By default, *mixing_ratio* refers to mass mixing ratios. The long name should
-   explicitly specify that it refers to the *mass* mixing ratio. 
-   Mass mixing ratios should contain information regarding 
-   with respect to what quantity they are defined, and options are *wrt_dry_air*, *wrt_moist_air*, 
+   explicitly specify that it refers to the *mass* mixing ratio.
+   Mass mixing ratios should contain information regarding
+   with respect to what quantity they are defined, and options are *wrt_dry_air*, *wrt_moist_air*,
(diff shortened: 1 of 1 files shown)


StandardNamesRules.rst
```diff
@@ -58,27 +58,33 @@ CCPP Standard Name Rules
...
+#. Number concentration should appear as a prefix, that is, *number_concentration_of*. By default,
    number concentrations are specified per unit of volume. When they are specified per
-   unit of mass, they should be written as *mass_number_concentration_of*. 
+   unit of mass, they should be written as *mass_number_concentration_of*.
+
+#. By default, *precipitation* refers to the sum of all phases of precipitating hydrometeors,
+   for example rain plus graupel plus hail.  The term *frozen_precipitation* refers to the
+   sum of all frozen precipitating hydrometers, for example graupel plus hail (but not rain).
+   Otherwise the standard name should explicitly state the type of hydrometeor(s) the
+   named quantity represents.
```
cacraigucar, line 87: Could the explanation give an example of what the name for a single precipitation should be (graupel_precipitation or precipitation_of_graupel).  I believe it could read `... named quanitity represents (i.e. graupel_precipitation)` if that is the correct one.

nusbaume: Sure, although I guess at that point the word `precipitation` might not be needed at all (e.g. you could just say `graupel`)?  It might be good to get people's thoughts on what that alternative name example should be either here or in the framework meeting, after which I'll be happy to update the rule with that example name.

mkavulich (approved):

cacraigucar (approved):

PR #38, opened 2023-05-31
Rename `specific_humidity`

### LINKED ISSUE #36: Non-universal definition of specific humidity
Variable `specific_humidity` is defined in rule 5 of the [CCPP Standard Name Rules](https://github.com/ESCOMP/CCPPStandardNames/blob/main/StandardNamesRules.rst).

>5.  By default, mixing_ratio refers to mass mixing ratios. The long name should explicitly specify that it refers to the mass mixing ratio. Mass mixing ratios should contain information regarding with respect to what quantity they are defined, and options are wrt_dry_air, wrt_moist_air, or wrt_total_mass, where moist_air refers to dry air plus vapor and total_mass refers to dry air plus vapor and hydrometeors. **A special case exists in the case of mixing ratio of vapor: the standard name specific_humidity should be used instead of mixing_ratio_of_vapor_wrt_moist_air.**

This definition as an alias of `mixing_ratio_of_vapor_wrt_moist_air` is not universal. At the UK Met Office for instance, what we call specific humidity is `mixing_ratio_of_vapor_wrt_total_mass`. This definition is used consistently across all our code, from NWP to climate. The definition here, `_wrt_moist_air`, is used in the UFS/GFS physics (according to @climbfuji). 

There seems to be no agreement on what specific_humidity really means. I would have liked the definition to be changed to the Met Office's one, but a better and more acceptable solution would be to simply remove this ambiguous name in the CCPP standard.

`water_vapor_mixing_ratio_wrt_dry_air` is already a CCPP name, I propose to add `water_vapor_mixing_ratio_wrt_moist_air` and `water_vapor_mixing_ratio_wrt_total_mass` as a replacement for `specific_humidity`. 

Any comment welcome.

MayeulDestouches:
# Description

This PR follows discussion in issue #36 about the ambiguous definition of specific humidity. Specific humidity is usually defined as the mixing ratio of the mass of water vapor to the mass of moist air, but moist air includes hydrometeors or not  depending on who you are talking to.

This PR modifies rule 5 on mixing ratios:
- The suffixes `_wrt_moist_air` and `_wrt_total_mass` are replaced by `wrt_moist_air` and `_wrt_moist_air_and_condensed_water`, which make it clearer that `moist_air` does not include condensed water in this standard. 
- `specific_humidity` is  no longer proposed as an alias for `water_vapor_mixing_ratio_wrt_moist_air`, not to favor this definition over `water_vapor_mixing_ratio_wrt_moist_air_and_condensed_water`.

This PR also modifies the standard names `specific_humidity` and `specific_humidity_at_2m` into their explicit versions `water_vapor_mixing_ratio_wrt_moist_air` [ `_and_condensed_water` ] [ `_at_2m` ].

# Issue addressed

- fixes #36 

# Notes

I haven't replaced the 22 other names that are built on variants `specific_humidity`. If required, I can replace them by the current CCPP definition which is `water_vapor_mixing_ratio_wrt_moist_air`.

### DIFF
--- a/Metadata-standard-names.md
+++ b/Metadata-standard-names.md
@@ -235,7 +235,9 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
     * `integer(kind=kind_phys)`: units = count
 * `number_of_tracers`: Number of tracers
     * `integer(kind=kind_phys)`: units = count
-* `specific_humidity`: Specific humidity
+* `water_vapor_mixing_ratio_wrt_moist_air`: Ratio of the mass of water vapor to the mass of moist air
+    * `real(kind=kind_phys)`: units = kg kg-1
+* `water_vapor_mixing_ratio_wrt_moist_air_and_condensed_water`: Ratio of the mass of water vapor to the mass of moist air and hydrometeors
     * `real(kind=kind_phys)`: units = kg kg-1
 * `mole_fraction_of_water_vapor`: Mole fraction of water vapor
     * `real(kind=kind_phys)`: units = mol mol-1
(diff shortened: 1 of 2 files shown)


grantfirl (approved):
This is OK with me.

nusbaume (approved):
This change is good on NCAR's end, and since all reviewers have approved I'll also go ahead and merge it as well.  Thanks for eliminating an extra source of name ambiguity!

PR #44, opened 2023-07-19
Adding atmospheric composition variables for JCSDA needs

jeromebarre:
Adding atmospheric composition variables for JCSDA needs.
I took the freedom to change constituent to atmospheric composition as this is a more generic term and is now more widely used in the scientific community,
I reordered this section as follows:
- variable counts
- water
- trace gas
- aerosols 

It would have been better to put them in XML subsections but the python script doesn't allow this. I could however update the python script in a different PR if we think the subsections are really needed. 

Note that there redundancies in certain quantities such as for O3 and CO2 as volume mixing ratio and mole fraction is the same. I understand that some models especially on the meteorology side have O3 and CO2 concentrations as mole fractions but the chemistry modules as volume mixing ratios. Both should be kept to satisfy the needs but we should try to limit the redundancies as much as possible.

### DIFF
--- a/standard_names.xml
+++ b/standard_names.xml
@@ -409,7 +409,7 @@
     <standard_name name="mole_fraction_of_ozone_in_air">
       <type kind="kind_phys" units="mol mol-1">real</type>
     </standard_name>
-    <standard_name name="mole_fraction_of_carbon_dioxide_in_air"
+    <standard_name name="mole_fraction_of_carbon_dioxide_in_air">
       <type kind="kind_phys" units="mol mol-1">real</type>
     </standard_name>
     <standard_name name="volume_mixing_ratio_of_ch4"
--- a/standard_names.xml
+++ b/standard_names.xml
@@ -425,7 +425,7 @@
       <type kind="kind_phys" units="mol mol-1">real</type>
     </standard_name>
     <standard_name name="volume_mixing_ratio_of_ccl4"
-                   long_name="CCL4 volume mixing ratio">
+                   long_name="Tetrachloromethane volume mixing ratio">
       <type kind="kind_phys" units="mol mol-1">real</type>
     </standard_name>
     <standard_name name="volume_mixing_ratio_of_cfc11"
--- a/standard_names.xml
+++ b/standard_names.xml
@@ -476,7 +476,8 @@
                    long_name="SO2 volume mixing ratio">
       <type kind="kind_phys" units="mol mol-1">real</type>
     </standard_name>
-
+  </section>
+    <section name="atmospheric_composition: GOCART aerosols">
     <standard_name name="mass_fraction_of_dust001_in_air"
                    long_name="Dust bin1 mass fraction">
       <type kind="kind_phys" units="kg kg-1">real</type>
--- a/standard_names.xml
+++ b/standard_names.xml
@@ -368,13 +368,15 @@
       <type kind="kind_phys" units="m s-1">real</type>
     </standard_name>
   </section>
-  <section name="constituents">
+  <section name="atmospheric composition">
+
     <standard_name name="number_of_chemical_species">
       <type kind="kind_phys" units="count">integer</type>
     </standard_name>
     <standard_name name="number_of_tracers">
       <type kind="kind_phys" units="count">integer</type>
     </standard_name>
+
     <standard_name name="water_vapor_mixing_ratio_wrt_moist_air"
                    long_name="Ratio of the mass of water vapor to the mass of moist air">
       <type kind="kind_phys" units="kg kg-1">real</type>
@@ -384,12 +386,6 @@
       <type kind="kind_phys" units="kg kg-1">real</type>
     </standard_name>
     <standard_name name="mole_fraction_of_water_vapor">
-      <type kind="kind_phys" units="mol mol-1">real</type>
-    </standard_name>
-    <standard_name name="mole_fraction_of_ozone_in_air">
-      <type kind="kind_phys" units="mol mol-1">real</type>
-    </standard_name>
-    <standard_name name="mole_fraction_of_carbon_dioxide_in_air">
       <type kind="kind_phys" units="mol mol-1">real</type>
     </standard_name>
     <standard_name name="water_vapor_mixing_ratio_wrt_dry_air"
@@ -411,6 +407,13 @@
     <standard_name name="rain_mixing_ratio_wrt_moist_air"
                    long_name="ratio of the mass of rain to the mass of moist air">
       <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+
+    <standard_name name="mole_fraction_of_ozone_in_air">
+      <type kind="kind_phys" units="mol mol-1">real</type>
+    </standard_name>
+    <standard_name name="mole_fraction_of_carbon_dioxide_in_air">
+      <type kind="kind_phys" units="mol mol-1">real</type>
     </standard_name>
     <standard_name name="volume_mixing_ratio_of_ch4"
                    long_name="CH4 volume mixing ratio">
@@ -452,6 +455,116 @@
                    long_name="N2O volume mixing ratio">
       <type kind="kind_phys" units="mol mol-1">real</type>
     </standard_name>
+    <standard_name name="volume_mixing_ratio_of_no2"
+                   long_name="NO2 volume mixing ratio">
+      <type kind="kind_phys" units="mol mol-1">real</type>
+    </standard_name>
+    <standard_name name="volume_mixing_ratio_of_no"
+                   long_name="NO volume mixing ratio">
+      <type kind="kind_phys" units="mol mol-1">real</type>
+    </standard_name>
+    <standard_name name="volume_mixing_ratio_of_o3"
+                   long_name="O3 volume mixing ratio">
+      <type kind="kind_phys" units="mol mol-1">real</type>
+    </standard_name>
+    <standard_name name="volume_mixing_ratio_of_hcho"
+                   long_name="HCHO volume mixing ratio">
+      <type kind="kind_phys" units="mol mol-1">real</type>
+    </standard_name>
+    <standard_name name="volume_mixing_ratio_of_c5h8"
+                   long_name="C5H8 volume mixing ratio">
+      <type kind="kind_phys" units="mol mol-1">real</type>
+    </standard_name>
+    <standard_name name="volume_mixing_ratio_of_so2"
+                   long_name="SO2 volume mixing ratio">
+      <type kind="kind_phys" units="mol mol-1">real</type>
+    </standard_name>
+
+    <standard_name name="mass_fraction_of_dust001_in_air"
+                   long_name="Dust bin1 mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_dust002_in_air"
+                   long_name="Dust bin2 mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_dust003_in_air"
+                   long_name="Dust bin3 mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_dust004_in_air"
+                   long_name="Dust bin4 mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_dust005_in_air"
+                   long_name="Dust bin5 mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_sea_salt001_in_air"
+                   long_name="Sea salt bin5 mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_sea_salt002_in_air"
+                   long_name="Sea salt bin2 mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_sea_salt003_in_air"
+                   long_name="Sea salt bin3 mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_sea_salt004_in_air"
+                   long_name="Sea salt bin4 mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_sea_salt005_in_air"
+                   long_name="Sea salt bin5 mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_hydrophobic_black_carbon_in_air"
+                   long_name="Hydrophobic black carbon mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_hydrophilic_black_carbon_in_air"
+                   long_name="Hydrophilic black carbon mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_hydrophobic_organic_carbon_in_air"
+                   long_name="Hydrophobic black carbon mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_hydrophilic_organic_carbon_in_air"
+                   long_name="Hydrophilic black carbon mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_sulfate_in_air"
+                   long_name="Sulfate mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_sea_nitratet001_in_air"
+                   long_name="Nitrate bin1 mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_sea_nitratet002_in_air"
+                   long_name="Nitrate bin2 mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_sea_nitratet003_in_air"
+                   long_name="Nitrate bin3 mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="volume_extinction_in_air_due_to_aerosol_particles_lambda1"
+                   long_name="Aerosol extinction at wavelength1">
+      <type kind="kind_phys" units="m-1">real</type>
+    </standard_name>
+    <standard_name name="volume_extinction_in_air_due_to_aerosol_particles_lambda2"
+                   long_name="Aerosol extinction at wavelength2">
+      <type kind="kind_phys" units="m-1">real</type>
+    </standard_name>
+    <standard_name name="volume_extinction_in_air_due_to_aerosol_particles_lambda3"
+                   long_name="Aerosol extinction at wavelength3">
+      <type kind="kind_phys" units="m-1">real</type>
+    </standard_name>
+
   </section>
   <section name="standard_variables"
            comment="Standard / required CCPP variables">
--- a/standard_names.xml
+++ b/standard_names.xml
@@ -368,7 +368,7 @@
       <type kind="kind_phys" units="m s-1">real</type>
     </standard_name>
   </section>
-  <section name="atmospheric composition">
+  <section name="atmospheric_composition">
 
     <standard_name name="number_of_chemical_species">
       <type kind="kind_phys" units="count">integer</type>
--- a/standard_names.xml
+++ b/standard_names.xml
@@ -369,14 +369,12 @@
     </standard_name>
   </section>
   <section name="atmospheric_composition">
-
     <standard_name name="number_of_chemical_species">
       <type kind="kind_phys" units="count">integer</type>
     </standard_name>
     <standard_name name="number_of_tracers">
       <type kind="kind_phys" units="count">integer</type>
     </standard_name>
-
     <standard_name name="water_vapor_mixing_ratio_wrt_moist_air"
                    long_name="Ratio of the mass of water vapor to the mass of moist air">
       <type kind="kind_phys" units="kg kg-1">real</type>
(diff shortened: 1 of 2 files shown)


gold2718:
> as volume mixing ratio and mole fraction is the same

Well, they are the same in an ideal gas. I understand that the difference is almost always `< 1%` in the atmosphere but I think they are in principle different physical quantities.

gold2718:
> I could however update the python script in a different PR if we think the subsections are really needed.

I am in favor of this. I would also love to see host-model-specific areas so that sections that are for interoperability would be separate from variables that are only used by one host model.

standard_names.xml
```diff
@@ -452,6 +455,116 @@
...
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_sea_salt004_in_air"
+                   long_name="Sea salt bin4 mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_sea_salt005_in_air"
+                   long_name="Sea salt bin5 mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
```
gold2718, line 522: Which set of bins is this? NGAC? ESA?
There are models and studies using different bin sizes as well as number of bins (e.g., more coarse bins to better capture the mass distribution in the atmosphere).
To be unambiguous, I think these should specify model of model type.
Maybe add `ngac` to these names?

standard_names.xml
```diff
@@ -452,6 +455,116 @@
...
+    <standard_name name="mass_fraction_of_hydrophilic_black_carbon_in_air"
+                   long_name="Hydrophilic black carbon mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_hydrophobic_organic_carbon_in_air"
+                   long_name="Hydrophobic black carbon mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_hydrophilic_organic_carbon_in_air"
+                   long_name="Hydrophilic black carbon mass fraction">
```
gold2718, line 536: This long name does not match the standard name. Fix?

standard_names.xml
```diff
@@ -452,6 +455,116 @@
...
+    <standard_name name="mass_fraction_of_hydrophobic_black_carbon_in_air"
+                   long_name="Hydrophobic black carbon mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_hydrophilic_black_carbon_in_air"
+                   long_name="Hydrophilic black carbon mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_hydrophobic_organic_carbon_in_air"
+                   long_name="Hydrophobic black carbon mass fraction">
```
gold2718, line 532: This long name does not match the standard name. Fix?

standard_names.xml
```diff
@@ -452,6 +455,116 @@
...
+    </standard_name>
+    <standard_name name="mass_fraction_of_hydrophilic_organic_carbon_in_air"
+                   long_name="Hydrophilic black carbon mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_sulfate_in_air"
+                   long_name="Sulfate mass fraction">
+      <type kind="kind_phys" units="kg kg-1">real</type>
+    </standard_name>
+    <standard_name name="mass_fraction_of_sea_nitratet001_in_air"
```
gold2718, line 543: I'm not sure what "nitratet" is. Well, in Norwegian, it is the single definite form of "nitrat" (salt of nitric acid) but I'm guessing that is not what this is.
If it is a typo, please fix it here and in the two bins below.

standard_names.xml
```diff
@@ -452,6 +455,116 @@
...
+                   long_name="Nitrate bin1 mass fraction">
```
gold2718, line 544: Should this be
```suggestion
                   long_name="Sea nitrate bin1 mass fraction">
```
?
If so, please fix it here and in the two bins below.

standard_names.xml
```diff
@@ -452,6 +455,116 @@
...
+      <type kind="kind_phys" units="m-1">real</type>
+    </standard_name>
+    <standard_name name="volume_extinction_in_air_due_to_aerosol_particles_lambda2"
+                   long_name="Aerosol extinction at wavelength2">
+      <type kind="kind_phys" units="m-1">real</type>
+    </standard_name>
+    <standard_name name="volume_extinction_in_air_due_to_aerosol_particles_lambda3"
+                   long_name="Aerosol extinction at wavelength3">
+      <type kind="kind_phys" units="m-1">real</type>
+    </standard_name>
```
gold2718, line 566: Again, are these universal wavelenths? If not, please attach a model name to each standard name.

standard_names.xml
```diff
@@ -452,6 +455,116 @@
...
+
```
gold2718, line 567: Is there a reason for this blank line?

gold2718 (changes requested):
I have a few questions and concerns.

standard_names.xml
jeromebarre: It's tied to a data assimilation methodology that GMAO is performing. It is not conditioned by one model in particular.

standard_names.xml
jeromebarre: No, it's just nitrate. There is no such thing as sea nitrate.

standard_names.xml
jeromebarre: It's a typo. I will correct this. Unnskyld!

standard_names.xml
jeromebarre: good catch thanks

standard_names.xml
jeromebarre: This is used with the GOCART aerosol scheme. But can be used in other aerosol schemes. To be honest, the names are already very long, is it really necessary?

standard_names.xml
```diff
@@ -368,13 +368,15 @@
...
-  <section name="constituents">
+  <section name="atmospheric_composition">
+
     <standard_name name="number_of_chemical_species">
       <type kind="kind_phys" units="count">integer</type>
     </standard_name>
     <standard_name name="number_of_tracers">
       <type kind="kind_phys" units="count">integer</type>
     </standard_name>
+
```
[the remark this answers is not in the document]
jeromebarre: will remove the blank lines

jeromebarre:
> > as volume mixing ratio and mole fraction is the same
> 
> Well, they are the same in an ideal gas. I understand that the difference is almost always `< 1%` in the atmosphere but I think they are in principle different physical quantities.

no, they are precisely the exact same quantities...

standard_names.xml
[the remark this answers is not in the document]
jeromebarre: actually, would the xml be a bit more human-readable if we put paces to separate the sections?

jeromebarre:
The concern about the aerosol scheme is valid. But I don't think we should add the scheme info in the names. I would suggest we create subsections where we specify the aerosol scheme in that case: atmospheric_composition > aerosols > GOCART > list of names. Would that be acceptable?

jeromebarre:
I have committed and pushed the changes, the commit appears on my fork but then it's not showing on this PR. Not sure if I am missing a command or something else?

standard_names.xml
jeromebarre: In the future, I would like them to have only one variable where the wavelength is another dimension, so the lambda part would be removed. I don't have control over this as it is dependent on GMAO folks working on this specific methodology to make changes in the JEDI code.

standard_names.xml
jeromebarre: for my own readability of the xml code... I have corrected this now.

standard_names.xml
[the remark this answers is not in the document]
gold2718: I am okay with any plan as long as it is consistent.

standard_names.xml
gold2718: > This is used with the GOCART aerosol scheme. But can be used in other aerosol schemes. To be honest, the names are already very long, is it really necessary?

To answer this question, you would have to know if anyone using that name would expect that the quantity is the same as in their model. People using a variable that ends up having a different physical definition than what they thought is one of the reasons the CCPP was invented in the first place.

standard_names.xml
gold2718: Takk, det går bra.

gold2718:
> > > as volume mixing ratio and mole fraction is the same
> > 
> > 
> > Well, they are the same in an ideal gas. I understand that the difference is almost always `< 1%` in the atmosphere but I think they are in principle different physical quantities.
> 
> no, they are precisely the exact same quantities...

I guess we took different chemistry courses :) My understanding is that they are the same as long as the volume
occupied by the gas is proportional to the number of molecules. I believe this assumption does not hold for gases such as C<sub>2</sub>H<sub>4</sub> or even CO<sub>2</sub> at high pressures.

standard_names.xml
gold2718: In the future, that extra dimension could be a coordinate so that the bands are defined by the coordinate variable but right now, those numbers could mean different things to different users. Can a case be made that no other system could ever have a use for these variables (a strong claim)?
Another idea (here and above with the GOCART bins) is that if we can create sections where model-specific standard names live, maybe we should allow this sort of name. Thoughts? @climbfuji? @grantfirl? @peverwhee?

gold2718:
> The concern about the aerosol scheme is valid. But I don't think we should add the scheme info in the names, they are already very long. I would suggest we create subsections where we specify the aerosol scheme in that case: atmospheric_composition > aerosols > GOCART > list of names. I think it is done somewhere where GFS is specified in the section title but not every variable has gfs in it. Would this solution be acceptable?

This is similar to what I just wrote above. However, currently, the GFS variables are not pulled out into a separate host-model section for UFS (and other models that use the same GFS physics structures?). If we (i.e. you :) _could_ create this sort of sectioning, I would be happy. If the group agrees with this plan, then I would suggest that you segregate them now and implement the sectioning in a new PR (as you originally proposed).

gold2718:
> I have committed and pushed the changes, the commit appears on my fork but then it's not showing on this PR. Not sure if I am missing a command or something else?

That's just GitHub taking a break. I see the changes now.

jeromebarre:
> > > > as volume mixing ratio and mole fraction is the same
> > > 
> > > 
> > > Well, they are the same in an ideal gas. I understand that the difference is almost always `< 1%` in the atmosphere but I think they are in principle different physical quantities.
> > 
> > 
> > no, they are precisely the exact same quantities...
> 
> I guess we took different chemistry courses :) My understanding is that they are the same as long as the volume occupied by the gas is proportional to the number of molecules. I believe this assumption does not hold for gases such as C2H4 or even CO2 at high pressures.

https://acmg.seas.harvard.edu/education/introduction-atmospheric-chemistry chapter 1 section 1.1 first line...

gold2718:
> chapter 1 section 1.1 first line

Well, sure, but if you read the *whole* paragraph, you find:

>  "... since the volume occupied by an ideal gas is proportional to the number of molecules. Pressures in the  atmosphere are sufficiently low that the ideal gas law is always obeyed to within 1%."

So, yes, they are roughly equivalent in the Earth's atmosphere (to 1% or so) but they are not the same physical quantity.
I suppose I can Google as well as the next guy (even if I was relegated to the little technical school down the street from Prof. Jacob).

Okay, that took a few minutes: From [Chemistry: The Central Science](https://chem.libretexts.org/Bookshelves/General_Chemistry/Map%3A_Chemistry_-_The_Central_Science_(Brown_et_al.))
by Brown, LeMay, Bursten, Murphy, and Woodward"
https://chem.libretexts.org/Bookshelves/General_Chemistry/Map%3A_Chemistry_-_The_Central_Science_(Brown_et_al.)/10%3A_Gases/10.09%3A_Real_Gases_-_Deviations_from_Ideal_Behavior

jeromebarre:
> > chapter 1 section 1.1 first line
> 
> Well, sure, but if you read the _whole_ paragraph, you find:
> 
> > "... since the volume occupied by an ideal gas is proportional to the number of molecules. Pressures in the  atmosphere are sufficiently low that the ideal gas law is always obeyed to within 1%."
> 
> So, yes, they are roughly equivalent in the Earth's atmosphere (to 1% or so) but they are not the same physical quantity. I suppose I can Google as well as the next guy (even if I was relegated to the little technical school down the street from Prof. Jacob).
> 
> Okay, that took a few minutes: From [Chemistry: The Central Science](https://chem.libretexts.org/Bookshelves/General_Chemistry/Map%3A_Chemistry_-_The_Central_Science_(Brown_et_al.)) by Brown, LeMay, Bursten, Murphy, and Woodward" https://chem.libretexts.org/Bookshelves/General_Chemistry/Map%3A_Chemistry_-_The_Central_Science_(Brown_et_al.)/10%3A_Gases/10.09%3A_Real_Gases_-_Deviations_from_Ideal_Behavior

In our context, atmospheric chemistry, is considered the same for the reason that we apparently both know. So the unit of the volume mixing ratio and the mole fraction is mol /mol and therefore is the same and this is redundant in **our context**. My point was to mention that those variables are redundant and that mole fraction and volume mixing ratio from the same model and the gas are very very likely to be bit identical... I think we should probably move on now...

gold2718:
>  our context

That's the thing, what is "our context"? For instance, our atmosphere model is also used to study different planets with quite different conditions so we try to not build unnecessary assumptions into the model. I would rather not build unnecessary assumptions into the CCPP if they could hinder future work.

jeromebarre:
ha! i didn't know that this was considering extraterrestrial atmospheres here... as long as you have mixing ratios in mol/mol,  which is used across the entire community this will be the same as a mole fraction. we can go on like this and bore to death everyone else on this PR...

cacraigucar:
I have removed myself as a reviewer as these are names which are outside my area of expertise.  I have requested @mattldawson as an alternate reviewer as this will overlap with the models that he works on.

climbfuji:
@jeromebarre @gold2718 What is the current status of this PR? Have you settled your discussion and agreed on a path forward?

jeromebarre:
Not really sure. I've corrected what was possible to correct in the XML file. For the long discussion with @gold2718  I was just pointing out that there might be redundant concentration definitions. This is valid only in earth's atmosphere but not in outer space... If this is used to run models of very high-pressure atmospheres (e.g. venus) the redundant quantities are not. Maybe @gold2718 can be more specific on what he now needs to see, but I am not sure what is still expected on this PR to be merged.

standard_names.xml
```diff
@@ -452,6 +452,115 @@
...
+      <type kind="kind_phys" units="mol mol-1">real</type>
+    </standard_name>
+    <standard_name name="volume_mixing_ratio_of_no"
+                   long_name="NO volume mixing ratio">
+      <type kind="kind_phys" units="mol mol-1">real</type>
+    </standard_name>
+    <standard_name name="volume_mixing_ratio_of_o3"
+                   long_name="O3 volume mixing ratio">
+      <type kind="kind_phys" units="mol mol-1">real</type>
+    </standard_name>
```
gold2718, line 466: I just realized that the [standard name rules](https://github.com/ESCOMP/CCPPStandardNames/blob/main/StandardNamesRules.rst) specify [`ozone`](https://github.com/ESCOMP/CCPPStandardNames/blob/main/StandardNamesRules.rst#chemical-species).
I guess that means this should be `volume_mixing_ratio_of_ozone`.

gold2718:
> @jeromebarre @gold2718 What is the current status of this PR? Have you settled your discussion and agreed on a path forward?

I guess it depends on whether we can reach agreement on future work on two fronts:

1. There are several variable sections that seem model specific to me (e.g., `mass_fraction_of_dustxxx_in_air` and `mass_fraction_of_sea_saltxxx_in_air`). If we can agree that they should be separated into a model-specific section (as proposed in #46), then I'm okay with not doing that in this PR.
2. We have a disagreement over the meaning of the word, 'redundant' (in my book numbers that differ by 1% are not redundant). However, at the top of this PR, @jeromebarre stated: "Both should be kept ..." so as long as that is true, that part of the discussion is not holding up this PR.

Of course, in trying to answer your question, I found another problem (`_o3` vs. `_ozone`) that I think we should resolve. Personally, I prefer `o3` but right now the rules disagree so I think we should either change `volume_mixing_ratio_of_o3` or remove ozone from the list of spelled-out species (and fix the 11 current standard names using `_ozone`).

standard_names.xml
jeromebarre: I think it makes more sense to have o3 for consistency with other molecules.

climbfuji: I'd be ok with using o3 throughout instead of ozone, and having something in the rules that specify what names to use for molecules (some way to encode the chemical formula in a word). @cacraigucar ?

cacraigucar: While I personally agree that this should be `o3` instead of `ozone`, I think we should also get votes from @nusbaume and more @mattldawson

mkavulich:
@jeromebarre We discussed the issues brought up in this PR at today's CCPP Framework meeting, and we proposed the following path forward:

1. Rename instances of `o3` to `ozone`. This is a standard from the CF conventions and is consistent with existing standard names.
2. Move the model-dependent variables to a new "GOCART" section. This may not be a permanent solution, but at least in the short term it will address the potential ambiguity for these variables.

The model-dependent variables (from what I can see) are 
 - mass_fraction_of_dust001_in_air
 - mass_fraction_of_dust002_in_air
 - mass_fraction_of_dust003_in_air
 - mass_fraction_of_dust004_in_air
 - mass_fraction_of_dust005_in_air
 - mass_fraction_of_sea_salt001_in_air
 - mass_fraction_of_sea_salt002_in_air
 - mass_fraction_of_sea_salt003_in_air
 - mass_fraction_of_sea_salt004_in_air
 - mass_fraction_of_sea_salt005_in_air
 - mass_fraction_of_sea_nitrate001_in_air
 - mass_fraction_of_sea_nitrate002_in_air
 - mass_fraction_of_sea_nitrate003_in_air
 - volume_extinction_in_air_due_to_aerosol_particles_lambda1
 - volume_extinction_in_air_due_to_aerosol_particles_lambda2
 - volume_extinction_in_air_due_to_aerosol_particles_lambda3

climbfuji:
@jeromebarre If you *really* want the CF conventions to change, then you need to reach out to them ... good luck. CCPP standard names need to follow those conventions unless we have a strong and good reason not to do so. Is this reason strong/good enough?

A compromise that I could imagine would be to say that the rule/preference is to use the common name of the molecule, but that aliases using the chemical formula (is there an ascii-standard notation for it?) are allowed. Such an automatic translation/match of names/alies could even be baked into the CCPP framework. Just an idea, needs further discussion.

nusbaume:
@jeromebarre With regards to your second point it looks like having sub-sections currently won't work.  However, I personally think that is a great idea.  If you want to try and tackle it yourself then feel free, but otherwise we can just make an issue and have one of us work on it in a separate PR, as that way this particular PR won't be held up.  Let me know if you have a preference.  Thanks!

jeromebarre:
> @jeromebarre With regards to your second point it looks like having sub-sections currently won't work. However, I personally think that is a great idea. If you want to try and tackle it yourself then feel free, but otherwise we can just make an issue and have one of us work on it in a separate PR, as that way this particular PR won't be held up. Let me know if you have a preference. Thanks!

Yes, I think we'd need to modify the Python script (I now remember this when reading the description I put for this PR). I can create sections that are placeholders for future subsections such as Composition - Trace Gas, Composition - Aerosols - GOCART, etc...

nusbaume:
@jeromebarre Just letting you know that I have created an issue (#48) for adding a subsection element.   Once we get that issue resolved we can hopefully move your sections into their respective subsections, assuming the names are still necessary.

mkavulich:
@jeromebarre just checking in to make sure you aren't waiting for action on our end. From what I can tell the path forward is:

1. Break out the model-dependent variables into a new section. We can re-arrange into subsections at a later date once #48 is addressed.
2. Rename instances of o3 to ozone for consistency with existing standard names and CF conventions. If we are going to discuss changing the names of chemical species that should be handled in another issue.

jeromebarre:
Yes sorry, I was on leave the last week and I have been busy with many other tasks lately. I will address this as soon as possible. Thanks for your patience. 

Note that, reverting the instances of o3 to ozone won't work for us at JCSDA right now. This will require us to change things in several areas of the JEDI code if we want to use that convention.  So we will create another inconsistency somewhere else and also with other trace gases' naming. Wouldn't we want to resolve the travel gas convention naming before merging this?

jeromebarre:
I've now added a section for GOCART aerosols. Unfortunately, this group must be aware that there would be only scheme-specific variables for aerosol in 3D Earth system modelling. GOCART is one of the schemes with a low number of variables comparatively. It is also the most widely used in the US in operations because of this. This scheme can be portend in various global and regional models so in a sense it is model-independent. Another scheme that is also used in our DA context at JCSDA is CMAQ which includes many more variables. So CCPP needs to prepare for this one way or another. Happy to help with the discussion for this. Also looking forward to the meeting to resolve the ozone/o3 problem. Thanks for your patience.

mkavulich:
Following a productive discussion today, we have agreed that this PR can go through as-is (with a few cleanup changes as proposed by @jeromebarre), and a future PR (I will open an issue to reference here) will do two things for cleanup, consistency, and looking forward to future names:

1. Convert existing mentions of `ozone` to `o3`, `carbon_dioxide` to `co2`, and any other chemical species names to their chemical symbols
2. For each existing chemical species mentioned by name in the standard names, we will add that chemical as a stand-alone standard name, with the long name for that standard name serving as an unambiguous definition for that chemical species that will carry over to other standard names.

We can discuss any changes to the above two points once I have opened that issue.

jeromebarre:
I've just updated the description of the volume mixing ratio to long names. I've left the mole fraction items as is.

standard_names.xml
```diff
@@ -412,46 +406,162 @@
...
       <type kind="kind_phys" units="mol mol-1">real</type>
     </standard_name>
     <standard_name name="volume_mixing_ratio_of_cfc12"
-                   long_name="CFC12 volume mixing ratio">
+                   long_name="Dichlorodifluoromethane volume mixing ratio">
       <type kind="kind_phys" units="mol mol-1">real</type>
     </standard_name>
     <standard_name name="volume_mixing_ratio_of_cfc113"
-                   long_name="CFC113 volume mixing ratio">
+                   long_name="1,1,2-Trichloro-1,2,2-trifluoroethane volume mixing ratio">
```
climbfuji, line 440: Wow.

climbfuji (approved):
I believe this matches what we discussed yesterday, thanks for making those changes.

mkavulich:
It looks like I don't have permission to merge while there are still requested changes. Can we get an approval from @gold2718? I am not sure if he is still actively working with the capgen effort.

cacraigucar:
I resolved the conversation on behalf of @gold2718 (since we discussed this in our meeting and came to agreement on the o3/ozone conundrum.  Unfortunately, that does not appear to have helped the merge blockage.  Hopefully @gold2718 sees this soon.

cacraigucar:
And then I hit the wrong button... sigh - I've reopened!

gold2718 (approved):

gold2718:
> It looks like I don't have permission to merge while there are still requested changes. Can we get an approval from @gold2718? I am not sure if he is still actively working with the capgen effort.

Sorry, I forgot I had not revisited this. One thing that really helps is to hit the little circular arrows icon in the reviewers box. This pings that reviewer to ask them to re-review their change requests.

mkavulich:
Thanks @gold2718, I hadn't used that functionality before so I'll keep that in mind for the future. Merging now.

PR #45, opened 2023-08-11
Updating gravity and fixing typos

### LINKED ISSUE #11: fix typos
x_wind: Horizontal wind in a direction perdendicular to y_wind

y_wind: Horizontal wind in a direction perdendicular to x_wind

"Perpendicular" is mistyped

Units for avogadro are "molecule mole-1", but will not match the current CAM definition (no change needed here)

### LINKED ISSUE #39: gravity and the "coordinates" section
Currently the standard name `gravitational_acceleration` is located under the `coordinates` section.  However,  I at least tend to not think of gravity as a coordinate, which then led to a discussion of what a "coordinate" actually means, at least in CCPP standard name world, which might be good to write down somewhere.  To start the discussion I can think of two possibilities for what constitutes a "coordinate":

1.  A quantity that is only a function or attribute of the model grid, and which is solely used to determine one's location in space.  With this definition arguably only `latitude` and `longitude` in the current dictionary file would classify as "coordinates".
2.  Any quantity that can be calculated using only geospatial information like the model grid and planet radius.  With this definition the `cell_area` and `cell_weight` would also be "coordinate" variables.

If we use definition 2 above then `gravitational_acceleration` could also stay given that is generally only a function of the distance from the planet's gravitational center.  However, one could also make the argument that if simulating non-Earth planets then the planet's mass also comes into play, which would then mean that `gravitational_acceleration` might be better included in a different section, like `state_variables`.

Finally, although in reality gravity varies as a function of one's location in space, many models treat it as a constant.  Would it also be beneficial to include a new standard name, say `standard_acceleration_of_gravity`, to represent this model constant?

Anyways, happy to hear any thoughts or opinions on this topic.  Thanks!

nusbaume:
Added a new "scalar" gravitational acceleration standard name, and move the variable  version into the `state_variables` section.  

I also fixed some typos and updated some other standard names as described in the `fix typos` issue.  I should note that one standard name was changed slightly more than what was described in the issue at a scientist's request. 

Fixes #39 
Fixes #11

### DIFF
--- a/Metadata-standard-names.md
+++ b/Metadata-standard-names.md
@@ -82,7 +82,7 @@ Currently, the only dimension which supports all six dimension types is horizont
     * `real(kind=kind_phys)`: units = kg m-3
 * `ratio_of_water_vapor_to_dry_air_gas_constants_minus_one`: (Rwv / Rdair) - 1.0
     * `real(kind=kind_phys)`: units = 1
-* `standard_gravitational_acceleration`: scalar constant epresenting gravitiational acceleration
+* `standard_gravitational_acceleration`: scalar constant representing gravitiational acceleration
     * `real(kind=kind_phys)`: units = m s-2
 ## coordinates
 * `latitude`: Latitude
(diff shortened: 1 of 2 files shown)


cacraigucar (approved):
It appears that the incorrect units in #11 were previously fixed

climbfuji (approved):

PR #47, opened 2023-08-23
Add CCPP metadata file standard name checking script

### LINKED ISSUE #41: CCPP metadata file standard name checking tool
It would be nice to have a script/tool that takes in the standard name dictionary and a CCPP metadata file (or a list of metadata files, or a path to a directory that contains metadata files) and checks for standard names in the metadata file(s) that are not in the dictionary.  The tool could then list all of those standard names not present in the  dictionary along with where they are located in the metadata file(s).

This tool is not specifically required for anything, but could be helpful for future physics scheme developers (or at least us at NCAR as we convert our physics to be CCPP-compliant).

nusbaume:
This Pull Request adds a new `meta_stdname_check.py` script that takes a path either to a metadata file or a directory and attempts to parse all of the found metadata files to see if there are any standard names that do not exist in the provided standard names dictionary, and then attempts to print the list to stdout in a hopefully easy-to-read way.

Along with this, I moved all of the other python files into a new `tools` and `lib` directory, to try and reduce the flatness of the repo.  This also required me to modify some of these scripts, as well as the Github Action workflow file, to make sure everything still works the way it is expected to.

Finally, I should note that I manually tested all of the scripts, but haven't implemented any sort of formal testing of the scripts themselves.  If this is desired then please let me know.

Fixes #41


nusbaume:
@cacraigucar I took you off the reviewer list because its basically all python code, but if you would still like to review this PR then just let me know.  Thanks!

nusbaume:
Apologies for the sudden flurry of commits, but of course I found a bug as soon as I opened the PR.  It should be resolved now, and so this PR is ready to review whenever.

tools/meta_stdname_check.py
```diff
@@ -0,0 +1,357 @@
...
+    #Add input arguments to be parsed:
+    parser.add_argument('--metafile-loc',
+                        metavar='<path to directory or file>',
+                        action='store', type=str,
+                        help="Location of metadata file(s)")
+
+    parser.add_argument('--stdname-dict',
+                        metavar='<path to file>',
+                        action='store', type=str,
+                        help="Location of standard name dictionary (XML file)")
```
mkavulich, line 81:
```suggestion
    parser.add_argument('-m', '--metafile-loc',
                        metavar='<path to directory or file>',
                        action='store', type=str,
                        help="Location of metadata file(s)")

    parser.add_argument('-s', '--stdname-dict',
                        metavar='<path to file>',
                        action='store', type=str,
                        help="Location of standard name dictionary (XML file)")
```

Just a personal preference suggestion, I'm a big fan of short option names.

tools/meta_stdname_check.py
```diff
@@ -0,0 +1,357 @@
...
+
+    NOTE:
+
+    The CCPP-framework has much more advanced parsers
+    that can extract this same info, but bringing them
+    into this repo would require many additional
+    supporting source files to be brought in as well.
+
+    However, if it is found that this simplified parser
+    is hitting too many edge cases then it might be wise
```
mkavulich, line 132: I'm curious what you mean by "edge cases" here?

tools/meta_stdname_check.py
```diff
@@ -0,0 +1,357 @@
...
+                comment_index = stdname_text.find("#")
+
+                #If comment exists, then remove
+                #it from the standard name text:
+                if comment_index != -1:
+                    stdname_text = stdname_text[:comment_index]
+                #End if
+            #End if
+
+            #Add stripped/trimmed text to the standardname set:
```
mkavulich, line 167: Should we add a check here for duplicate standard names? Might be beyond the scope but would be a useful check (and I think easy to implement), especially as the list gets longer and such an addition might get overlooked.

mkavulich (approved):
Looks good, just some optional comments. Funnily enough, I randomly chose a CCPP physics meta file to test this new script (`ufs-weather-model/FV3/ccpp/physics/physics/rrtmgp_sw_main.meta`), and it has **a lot** of non-standard names. I was actually looking for a case that had all valid standard names for comparison and couldn't find one 😬

tools/meta_stdname_check.py
nusbaume: Good idea!  The suggested short option names have now been added.

tools/meta_stdname_check.py
nusbaume: I was thinking of metadata file text that trips up my simplified parser, but which is properly managed by the official CCPP-framework metadata file parser.  I am not personally aware of what any of those edge cases might be,  but just wanted to add this comment in case future users or developers are running into problems and want an idea of how to fix it without writing a bunch of new code.

tools/meta_stdname_check.py
nusbaume: This is probably not be what you are asking, but in python the "set" datatype automatically remove any duplicates, so the script should always report just one instance of each "missing" standard name even if it is repeated multiple times in the metadata file itself.

In terms of actually checking if a specific scheme phase has input or output arguments that have the same standard name (and thus are duplicates) then we could certainly modify this script to do so, but we would first need to parse for the specific scheme phase, as otherwise we'll hit "duplicate" names even if the metadata file is completely fine, simply because multiple different phases might need to use the same variable (e.g. `_timestep_init` and `_timestep_final` might both use `air_temperature`).

Anyways, hopefully that all makes sense?  I'm happy to discuss this further, especially if folks think checking for true duplicate arguments in this script would be useful (or if the framework itself should handle it).

tools/meta_stdname_check.py
dustinswales: I don't see anything wrong with this approach, but yeah, I agree that it would be better to pull in the ccpp-framework and use the metadata parsers in there. We can add it to the list (e.g. open an issue to remind us)?

dustinswales (commented):
@nusbaume Thanks for creating this script. Looks good to me.
As for using this for testing, I think it would be deployed outside of this repository? For example, in the ccpp-physics repositories and host-model repositories there could be CI tests to run this script and compare their metadata to the standard_name repository?

nusbaume:
Just FYI that I added a timestamp to the printed output as requested by @cacraigucar and @peverwhee.

nusbaume:
@dustinswales Apologies for missing your post!  Yes, I was assuming this would be a tool one could apply to CCPP metadata files in external repos that contain such files (e.g. `NCAR/ccpp-physics` or `NCAR/atmospheric_physics`). 

I certainly think that long-term it would be good to implement something like this script in a CI test to ensure that all standard names listed in a new metadata file are in the dictionary.  However, I haven't really pushed for it at this stage simply because I believe a lot of physics schemes don't have fully dictionary-compliant standard names, and that it will probably take some work before a critical mass of them do.

dustinswales:
@nusbaume No worries (I forgot that I made a comment...)
Yeah, we don't need to add this to our actions right away, but since you created it, we should put it to use at somepoint.

tools/meta_stdname_check.py
```diff
@@ -0,0 +1,363 @@
...
+The tool currently has two options:
+
+1.  A path to a single metadata file
+    is passed, in which case only that
+    file's standard names are checked, e.g.:
+
+./meta_stdname_check --metafile-loc /path/to/file.meta --stdname-dict /path/to/dict.xml
+
+2.  A path to a directory is passed, in
+    which case the directory is searched
```
peverwhee, line 19: Maybe change to:

> "the directory (and any subdirectories) is searched"

Not great grammar, but makes it clear that you can run this script on a top-level directory and still pick up all the metadata files.

peverwhee (approved):
One comment that you can take or leave.

Thanks Jesse!

tools/meta_stdname_check.py
nusbaume: Added text to make it clear that the subdirectories are searched as well.  Thanks for the suggestion!

Issue #50, opened 2023-09-29
Add suffix to indicate a 3D field has an additional point above the model top


ss421:
As part of the JEDI processing system, some of the Met Office operators use a vertical grid where there is an extra point added to include a point just above the model top (a full grid cell above the centre of the previous). This extra point is added just for the height-based fields like pressure and geometric height. So we have the following as an example:

```
air_pressure_at_interafce – on the interface (n+1 points)
air_pressure – cell centered (n points)
air_pressure_including_point_above_model_top – cell centered with an additional point above the model (n+1 points)
```

Since JEDI is adopting the CCPP standard, we need some way to distinguish these fields. We are proposing to add the `including_point_above_model_top` suffix.

climbfuji:
I think there is already a standard name rule or more, depending on the use case, for this. The simplest example is the suffix `_plus_one`. There are also other qualifiers that can be used to distinguish, for example `_for_radiation` (in case the vertical levels for radiation are more or less than for other fields) and the distinction between levels and interfaces (half levels, full levels). Do you any of this will work for you?

ss421:
Thankyou for the suggestion, I did not spot the `_plus one`. I see that `_plus_one` is defined in [StandardNamesRules](https://github.com/ESCOMP/CCPPStandardNames/blob/main/StandardNamesRules.rst) and used in [Metadata-standard-names](https://github.com/ESCOMP/CCPPStandardNames/blob/main/Metadata-standard-names.md) for: `number_of_tracers_plus_one`.

At the moment, we use `_levels` to distinguish between the two grid staggers. For the extra level we use the suffix `_minus_one` rather than `_plus_one`. I recall `_plus_one` was ruled out in favor of `_minus_one` but I was not part of the discussion in that case.

In terms of the new name, we wanted something descriptive and so opted for `_including_point_above_model_top`. I think `_plus_one` is a little ambiguous but I can ask others who have more of an opinion on this matter than me to contribute. This was discussed in [jcsda-interanl/vader#109](https://github.com/JCSDA-internal/vader/issues/109) (apologies to those that do not have access).

For reference, we have
```
current => will become
air_pressure => air_pressure_at_interface
air_pressure_levels => air_pressure_including_point_above_model_top
air_pressure_levels_minus_one => air_pressure
```

climbfuji:
Ok, thanks. Let's see what the other CCPP std name folks have to say.

gold2718:
[...]

PR #51, opened 2023-09-29
Add including_point_above_model_top --> extended_up_by_1 suffix

ss421:
### Description

~~In issue https://github.com/ESCOMP/CCPPStandardNames/issues/50, the requirement to have a suffix that indicates an additional point above the model top was outlined. This PR adds the suggested suffix to the rules document and adds the `air_pressure_including_point_above_model_top` variable that uses it to the standard_names.~~

~~Im adding this as a draft PR initially for feedback on the suitability of adding such a suffix.~~

**_Updated with feedback from_** #50.  The requirement to have a suffix that indicates an additional point above the model top was outlined and discussed. This PR adds the suggested suffix (`extended_up_by_1`) to the rules document and adds the `air_pressure_extended_up_by_1` variable and `vertical_layer_dimension_extended_up_by_1` dimension to the standard_names.

### Issue addressed

https://github.com/ESCOMP/CCPPStandardNames/issues/50

### DIFF
--- a/Metadata-standard-names.md
+++ b/Metadata-standard-names.md
@@ -162,6 +162,8 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
     * `real(kind=kind_phys)`: units = 1
 * `ln_air_pressure_of_dry_air_at_interface`: Ln air pressure of dry air at interface
     * `real(kind=kind_phys)`: units = 1
+* `air_pressure_including_point_above_model_top`: Air pressure including point above model top
+    * `real(kind=kind_phys)`: units = Pa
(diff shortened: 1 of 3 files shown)


nusbaume (approved):

gold2718 (approved):
I still think this naming scheme is both overly restrictive and redundant. Fixing the number one will just lead to needing more names (e.g., radiation which can have a variable number _could_ have used a more general syntax). Also, "up by 1" relative to what? The description says "vertical layers" so the source is `vertical_layer_dimension` and, as they say, just do the math.

Issue #42, opened 2023-06-13
Want Github Action to ensure standard names are following character rules


nusbaume:
As described by the CF-conventions here:

http://cfconventions.org/Data/cf-standard-names/docs/guidelines.html

standard names should start with a letter, and only contain lower-case letters, numbers, and underscores.  This rule should be enforceable in an automated fashion, and thus it would be good to have a Github Action for the dictionary that ensures that only standard names that follow this rule are allowed.

mkavulich:
Resolved via #52

PR #52, opened 2023-10-19
Fix invalid standard names, add standard name checking script and github action

mkavulich:
### Description

This PR introduces a Github Action to automatically check the validity of Standard Names against the list of criteria specified by the CF conventions: Must only contain lowercase letters, numerals, or underscores, and the first character must be a letter. This is achieved with a new script (`tools/check_name_rules.py`) that parses the given standard names xml file, and checks each name against these rules, returning an error if any invalid names are found. This action will run on all branches for any new commits to ensure any new names follow these rules.

This new script revealed several existing invalid names; those have been corrected as part of this PR.

### Issue addressed

#42


tools/check_name_rules.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python3
+
+"""
+Remove duplicates from a metadata standard-name XML library file.
+"""
+
+import argparse
+import sys
+import os.path
+import copy
```
nusbaume, line 10: I don't think `copy` is used in this script.

tools/check_name_rules.py
```diff
@@ -0,0 +1,80 @@
...
+    schema_file = find_schema_file(schema_path, version)
+    if schema_file:
+        try:
+            validate_xml_file(stdname_file, schema_name, version, None,
+                            schema_path=schema_root, error_on_noxmllint=True)
+        except ValueError:
+            raise ValueError("Invalid standard names file, {}".format(stdname_file))
+    else:
+        raise ValueError(
+            'Cannot find schema file, {}, for version {}'.format(schema_name, version)
```
nusbaume, line 56: Should we make these f-strings instead of using `format`?

tools/check_name_rules.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+
+    #Parse list of standard names and see if any names violate one or more rules
+    violators = {}
+    for name in root.findall('./section/standard_name'):
```
nusbaume, line 61: I might compile the regular expressions before the standard name loop:

```suggestion
    violators = {}
    legal_first_char = re.compile('[a-z]')
    valid_chars = re.compile('[a-z0-9_]')
    for name in root.findall('./section/standard_name'):
```

This should make the script slightly more performant (which might matter if the standard names dictionary gets huge).

tools/check_name_rules.py
```diff
@@ -0,0 +1,80 @@
...
+        sname = name.attrib['name']
+        violations = []
+        if re.sub('[a-z]', '', sname[0]):
```
nusbaume, line 64: If you compiled the regex above, then this becomes:

```suggestion
        if legal_first_char.sub('', sname[0]):
```

tools/check_name_rules.py
```diff
@@ -0,0 +1,80 @@
...
+            violations.append('First character is not a lowercase letter')
+        testchars = re.sub('[a-z0-9_]', '', sname)
```
nusbaume, line 66: If you compiled the regex above, then this becomes:

```suggestion
        testchars = valid_chars.sub('', sname)
```

mkavulich:
@nusbaume Thanks for the great review! I applied most of your suggestions, and also applied the analogous improvements to `tools/check_xml_unique.py`. 

Regarding the CI run rules, I thought it would be useful to always check the names so problems get caught at an early stage, but you're right that does impose a lot on users with forks. I changed the rule to pull-request-only, and since we already have a file for pull-request CI, I combined the new rule into the existing file.

Now, for some reason after all these updates the markdown check CI test is failing. I can't figure out what's going on here, as there's no difference locally between the two files.

nusbaume (approved):
Thanks for resolving my concerns!  I added a few more commits to fix the test failure and some syntax errors, so it should be good now (at least in my opinion).  Thanks again for taking this on!

climbfuji (approved):
Thanks for this, very helpful!

grantfirl (approved):

PR #55, opened 2023-12-01
Add function to apply conversion rules to generated markdown links

### LINKED ISSUE #54: Need to apply text-to-link conversion rules in Markdown generation
Github applies a set of text conversion rules to markdown header text when converting it into a link.

The markdown generation script converts standard name dictionary sections into markdown headers and header links, but doesn't apply these conversion rules, which can result in broken links.   Thus the `write_standard_name_table.py` script should be modified to apply those rules when writing the markdown file.

nusbaume:
This PR adds a function to apply header text conversion rules when generating internal markdown document links.

Fixes #54

### DIFF
--- a/tools/write_standard_name_table.py
+++ b/tools/write_standard_name_table.py
@@ -26,8 +26,38 @@ from xml_tools import validate_xml_file, read_xml_file
 from xml_tools import find_schema_file, find_schema_version
 
 #######################################
+#Regular expressions
+#######################################
 
 _REAL_SUBST_RE = re.compile(r"(.*\d)p(\d.*)")
+
+_DROPPED_LINK_CHARS_RE = re.compile(r"[^a-z_-]")
+
+########################################################################
+def convert_text_to_link(text_str):
+
+    """
+    When Markdown converts a header string into
+    an internal document link it applies certain
+    text conversion rules.  This function thus
+    applies those same rules to a given string
+    in order to produce the correct link.
+    """
+
+    #First trim the string to remove leading/trailing white space:
+    link_str = text_str.strip()
+
+    #Next, make sure all text is lowercase:
+    link_str = link_str.lower() #all text is lowercase
+
+    #Then, replace all spaces with dashes:
+    link_str = link_str.replace(" ", "-")
+
+    #Finally, remove all characters that aren't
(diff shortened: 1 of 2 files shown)


nusbaume:
Please note that these test failures should be fixed once PR #52 has been accepted and merged (both into main and my branch).

tools/write_standard_name_table.py
```diff
@@ -25,10 +25,40 @@
...
+    text conversion rules.  This function thus
+    applies those same rules to a given string
+    in order to produce the correct link.
+    """
+
+    #First trim the string to remove leading/trailing white space:
+    link_str = text_str.strip()
+
+    #Next, make sure all text is lowercase:
+    link_str = link_str.lower()
```
mkavulich, line 51: Is this check necessary? It appears as if the old links with capital letters still work in the web view, but maybe there's another markdown reader where it doesn't work?

mkavulich (approved):
Looks good! Just had one question.

tools/write_standard_name_table.py
nusbaume: That's a good question!  After doing some experimentation it looks like at least my web browser (Chrome) is able to use a URL with uppercase text correctly, but I think "under the hood" the actual link is still all lowercase.  You can see this if you go to to one of the section heads in the markdown document itself and then click on the the little chain symbol that pops up next to it.

So, I am happy to remove this text conversion if that is preferred by you or others, but I personally think keeping everything lowercase will be the safest option in the long run.  Of course please let me know if you disagree!

nusbaume:
@mkavulich I removed the "needs discussion" label on this PR because I _think_ you meant to add it to PR #56 instead based off of my memory of Monday's framework meeting discussions.  However, if I am wrong here then feel free to add the label back.  Thanks!

mkavulich:
@nusbaume Yes, you're right I applied the label to the wrong PR. Thanks for fixing that!

tools/write_standard_name_table.py
mkavulich: It's fine to leave the check in, I was just curious if it was causing issues.

grantfirl (approved):

PR #56, opened 2023-12-05
Add new standard names from NCAR idealized physics schemes

nusbaume:
This PR adds some new CCPP standard names that are currently being used within SIMA for NCAR idealized physics schemes.  

Along with the standard names themselves, I also added a couple extra sections for some variables that I thought wouldn't fit very nicely elsewhere.  At some point I think it would be worth discussing the organization of the standard names a little more, as there are some standard names that are listed under specific GFS DDTs but which are probably more universal than that (e.g. `mpi_rank`).  However, I don't believe this potential organizational concern needs to be resolved for this particular PR.

Finally, please note that some of the links to the new sections in the generated markdown file will be broken until PR #55 has been merged.  Thanks!

### DIFF
--- a/Metadata-standard-names.md
+++ b/Metadata-standard-names.md
@@ -7,7 +7,9 @@
 * [diagnostics](#diagnostics)
 * [atmospheric_composition](#atmospheric_composition)
 * [atmospheric_composition: GOCART aerosols](#atmospheric_composition: GOCART aerosols)
-* [standard_variables](#standard_variables)
+* [required standard variables](#required standard variables)
+* [optional standard variables](#optional standard variables)
+* [system variables](#system variables)
 * [GFS_typedefs_GFS_control_type](#GFS_typedefs_GFS_control_type)
 * [GFS_typedefs_GFS_interstitial_type](#GFS_typedefs_GFS_interstitial_type)
 * [GFS_typedefs_GFS_tbd_type](#GFS_typedefs_GFS_tbd_type)
@@ -130,6 +132,8 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
     * `logical(kind=)`: units = flag
 * `lagrangian_tendency_of_air_pressure`: Vertical pressure velocity
     * `real(kind=kind_phys)`: units = Pa s-1
+* `density_of_dry_air`: Density of dry air
+    * `real(kind=kind_phys)`: units = kg m-3
 * `air_pressure`: Midpoint air pressure
     * `real(kind=kind_phys)`: units = Pa
 * `air_pressure_of_dry_air`: Dry midpoint pressure
@@ -152,6 +156,8 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
     * `real(kind=kind_phys)`: units = m
 * `geopotential_height_wrt_surface`: geopotential height w.r.t. local surface
     * `real(kind=kind_phys)`: units = m
+* `geopotential_height_wrt_surface_at_interface`: geopotential height w.r.t. local surface at interface
+    * `real(kind=kind_phys)`: units = m
 * `potentially_advected_quantities`: Potentially advected quantities
     * `real(kind=kind_phys)`: units = various
 * `air_pressure_at_interface`: Air pressure at interface
@@ -168,6 +174,8 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
     * `logical(kind=kind_phys)`: units = flag
 * `is_initialized_physics_grid`: Flag to indicate if physics grid is initialized
     * `logical(kind=kind_phys)`: units = flag
+* `print_qneg_warn`: Logging setting for negative constituent mass fixer
+    * `character(kind=len=*)`: units = 1
 * `geopotential_height_at_interface`: Geopotential height at interface
     * `real(kind=kind_phys)`: units = m
 * `vertically_integrated_total_energy_of_initial_state`: Vertically integrated total energy of initial state
@@ -178,9 +186,11 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
     * `real(kind=kind_phys)`: units = kg m-2
 * `vertically_integrated_total_water_of_current_state`: Vertically integrated total water of current state
     * `real(kind=kind_phys)`: units = kg m-2
+* `heating_rate`: heating rate
+    * `real(kind=kind_phys)`: units = J kg-1 s-1
 * `tendency_of_air_temperature`: Change in temperature from a parameterization
     * `real(kind=kind_phys)`: units = K s-1
-* `tendency_of_air_temperature_due_to_model_physics`: Total change in temperature from a                               physics suite
+* `tendency_of_air_temperature_due_to_model_physics`: Total change in temperature from a physics suite
     * `real(kind=kind_phys)`: units = K s-1
 * `tendency_of_air_potential_temperature`: Change in potential temperature from a parameterization
     * `real(kind=kind_phys)`: units = K s-1
@@ -193,6 +203,14 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
 * `tendency_of_y_wind`: Change in y wind from a parameterization
     * `real(kind=kind_phys)`: units = m s-2
 * `tendency_of_y_wind_due_to_model_physics`: Tendency of y wind due to model physics
+    * `real(kind=kind_phys)`: units = m s-2
+* `tendency_of_eastward_wind`: Change in eastward wind from a parameterization
+    * `real(kind=kind_phys)`: units = m s-2
+* `tendency_of_eastward_wind_due_to_model_physics`: Total change in eastward wind from a physics suite
+    * `real(kind=kind_phys)`: units = m s-2
+* `tendency_of_northward_wind`: Change in northward wind from a parameterization
+    * `real(kind=kind_phys)`: units = m s-2
+* `tendency_of_northward_wind_due_to_model_physics`: Total change in northward wind from a physics suite
     * `real(kind=kind_phys)`: units = m s-2
 * `surface_upward_heat_flux_in_air`: Surface upward heat flux in air
     * `real(kind=kind_phys)`: units = W m-2
@@ -208,6 +226,8 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
     * `real(kind=kind_phys)`: units = Pa
 * `reference_air_pressure_normalized_by_surface_air_pressure`: reference pressure normalized by surface pressure
     * `real(kind=kind_phys)`: units = 1
+* `reference_pressure_in_atmosphere_layer_normalized_by_reference_pressure`: reference pressure in atmosphere layer normalized by reference pressure
+    * `real(kind=kind_phys)`: units = 1
 * `dimensionless_exner_function`: exner function
     * `real(kind=kind_phys)`: units = 1
 * `air_potential_temperature`: air potential temperature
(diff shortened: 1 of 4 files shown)


Metadata-standard-names.md
```diff
@@ -7,7 +7,9 @@
 * [diagnostics](#diagnostics)
 * [atmospheric_composition](#atmospheric_composition)
 * [atmospheric_composition: GOCART aerosols](#atmospheric_composition: GOCART aerosols)
-* [standard_variables](#standard_variables)
+* [required standard variables](#required standard variables)
+* [optional standard variables](#optional standard variables)
+* [system variables](#system variables)
```
mkavulich, line 12: I see that these links are currently broken, I assume that will get fixed once #55 is merged?

standard_names.xml
```diff
@@ -271,13 +283,16 @@
     <standard_name name="vertically_integrated_total_water_of_current_state">
       <type kind="kind_phys" units="kg m-2">real</type>
     </standard_name>
+    <standard_name name="heating_rate"
```
mkavulich, line 286: This variable name feels very vague, could it be made more specific?

If the standard name needs to be kept at is, it definitely needs a more specific `long_name` entry at the very least.

standard_names.xml
```diff
@@ -271,13 +283,16 @@
...
+      <type kind="kind_phys" units="J kg-1 s-1">real</type>
+    </standard_name>
     <standard_name name="tendency_of_air_temperature"
                    long_name="Change in temperature from a parameterization">
       <type kind="kind_phys" units="K s-1">real</type>
     </standard_name>
     <standard_name name="tendency_of_air_temperature_due_to_model_physics"
-                   long_name="Total change in temperature from a
-                              physics suite">
+                   long_name="Total change in temperature from a physics suite">
```
mkavulich, line 295: While we're changing things, could you make this addition?
```suggestion
                   long_name="Total change in air temperature from a physics suite">
```

standard_names.xml
```diff
@@ -579,6 +634,40 @@
       <type kind="" units="1">integer</type>
     </standard_name>
   </section>
+  <section name="optional standard variables"
+           comment="Optional CCPP variables">
+    <standard_name name="scheme_name"
+                   long_name="CCPP physics scheme name">
+      <type kind="len=64" units="none">character</type>
+    </standard_name>
+    <standard_name name="ccpp_constituent_properties"
```
mkavulich, line 643: Out of curiosity, can you point me to how these constituent variables are being used in SIMA? I'm curious about the specifics of the use case (for example, why there are minimum but not maximum values).

mkavulich (changes requested):

Metadata-standard-names.md
grantfirl: I personally don't like the term "standard variables" because "standard" is sort of a loaded term. It can mean normal/usual as it is meant here, but also referring to some standard. We're also using standard names to denote "the key by which a specific quantity is known by physics and hosts using the CCPP", and someone might draw some link between the two. What is really in this section are variables provided by the framework, right? So, could we change the section name to be "framework-provided", perhaps with separate sections for required framework-provided and optional framework-provided?

Metadata-standard-names.md
```diff
@@ -168,6 +174,8 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
     * `logical(kind=kind_phys)`: units = flag
 * `is_initialized_physics_grid`: Flag to indicate if physics grid is initialized
     * `logical(kind=kind_phys)`: units = flag
+* `print_qneg_warn`: Logging setting for negative constituent mass fixer
```
cacraigucar, line 177: I'm wondering if this should be `print_qneg_warning` instead?

Metadata-standard-names.md
```diff
@@ -130,6 +132,8 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
     * `logical(kind=)`: units = flag
 * `lagrangian_tendency_of_air_pressure`: Vertical pressure velocity
     * `real(kind=kind_phys)`: units = Pa s-1
+* `density_of_dry_air`: Density of dry air
```
grantfirl, line 135: CF conventions use "X_density" rather than "density_of_X".

Metadata-standard-names.md
```diff
@@ -246,14 +270,20 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
     * `real(kind=kind_phys)`: units = mol mol-1
 * `water_vapor_mixing_ratio_wrt_dry_air`: Ratio of the mass of water vapor to the mass of dry air
     * `real(kind=kind_phys)`: units = kg kg-1
+* `cloud_liquid_water_mixing_ratio_wrt_moist_air_and_condensed_water`: Ratio of the mass of liquid water to the mass of moist air and condensed water
```
cacraigucar, line 273: I'm not finding this in our local spreadsheet

cacraigucar (changes requested):

Metadata-standard-names.md
```diff
@@ -178,9 +186,11 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
     * `real(kind=kind_phys)`: units = kg m-2
 * `vertically_integrated_total_water_of_current_state`: Vertically integrated total water of current state
     * `real(kind=kind_phys)`: units = kg m-2
+* `heating_rate`: heating rate
```
grantfirl, line 189: How is this related to "tendency_of_air_temperature"? Does it refer to the heating of anything in particular?

Metadata-standard-names.md
```diff
@@ -208,6 +226,8 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
     * `real(kind=kind_phys)`: units = Pa
 * `reference_air_pressure_normalized_by_surface_air_pressure`: reference pressure normalized by surface pressure
     * `real(kind=kind_phys)`: units = 1
+* `reference_pressure_in_atmosphere_layer_normalized_by_reference_pressure`: reference pressure in atmosphere layer normalized by reference pressure
```
grantfirl, line 229: This is confusing to me. I'm guessing that the term "reference" means different things both times it is used, otherwise this would just be 1. Is there a better term than "reference" for either the first or second instance?

Metadata-standard-names.md
```diff
@@ -333,12 +363,30 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
...
 Standard / required CCPP variables
 * `ccpp_error_message`: Error message for error handling in CCPP
     * `character(kind=len=512)`: units = none
 * `ccpp_error_code`: Error code for error handling in CCPP
     * `integer(kind=)`: units = 1
+## optional standard variables
+Optional CCPP variables
+* `scheme_name`: CCPP physics scheme name
+    * `character(kind=len=64)`: units = none
+* `ccpp_constituent_properties`: CCPP Constituent Properties
```
grantfirl, line 376: Do we need to have "CCPP" in the standard names added here? Aren't all standard names "CCPP" standard names in this context?

Metadata-standard-names.md
grantfirl: Agreed. Also see rule 14: `For control-oriented variables, if the variable is a Fortran logical, use flag_for _X. If it is any other data type, use control_for _X. All flags should be Fortran logicals.` So, would `control_for_negative_constituent_warning` work? I realize that this is much longer than you have, though.

Metadata-standard-names.md
cacraigucar: I think at least this particular name should have ccpp in its name, as it will be the CCPP derived data type.   I don't have a strong feeling for the other variables listed below this one.

grantfirl: Maybe it's OK to keep `ccpp` in the standard name and maybe add a rule that all variables that start with `ccpp_X` are provided by the framework.

Metadata-standard-names.md
nusbaume: @mkavulich - Yes!  In fact I just merged in that PR and updated this branch, so the links should be fixed now.  Of course please let me know if they still don't work properly for you.

@grantfirl - This is a good point.  I went ahead and changed these sections to be `framework-provided variables` as suggested in order to avoid the overuse of the word "standard".

standard_names.xml
nusbaume: Done!

standard_names.xml
nusbaume: Sure!  You can find an example of us using it to calculate geopotential here:  https://github.com/ESCOMP/atmospheric_physics/blob/main/utilities/geopotential_temp.F90  

with the associated metadata file here:  https://github.com/ESCOMP/atmospheric_physics/blob/main/utilities/geopotential_temp.meta).  

We also use them for our negative constituent adjuster here:  https://github.com/ESCOMP/atmospheric_physics/blob/main/utilities/qneg.F90

with the associated metadata file here:  https://github.com/ESCOMP/atmospheric_physics/blob/main/utilities/qneg.meta)

If you have any questions or concerns about these routines or the general use of these data structures just let me know!

Metadata-standard-names.md
nusbaume: Good catch!  Standard name changed to `dry_air_density`.

Metadata-standard-names.md
nusbaume: @cacraigucar @grantfirl Thanks for catching this!  Went ahead and changed it to `control_for_negative_constituent_warning` to best match the naming rules.

Metadata-standard-names.md
nusbaume: Thanks for checking!  The name in the spreadsheet was wrong, but has been fixed now.

Metadata-standard-names.md
nusbaume: See the explanation I gave Mike here:  https://github.com/ESCOMP/CCPPStandardNames/pull/56#discussion_r1457612371

Happy to change the name if one of the alternative names is better (I can also update the long name if preferred).

Metadata-standard-names.md
nusbaume: Yeah, the issue is that this is basically the variable with the accepted standard name `reference_pressure_in_atmosphere_layer`, which is a 1-D array, divided by the variable with the accepted standard name `reference_pressure`, which is a scalar.

If I remember correctly there were possibly different terms for this quantity.  @PeterHjortLauritzen  @adamrher @JulioTBacmeister do any of you remember?  To help jog your memory this quantity is the same as `pref_mid_norm` in `ref_pres.F90` in CAM.

Metadata-standard-names.md
nusbaume: I have updated the rules to make it clear that any standard name that starts with `ccpp_` should be a framework-provided variable.

Metadata-standard-names.md
nusbaume: After chatting with folks I think everyone is ok with changing `reference_pressure` to `surface_reference_pressure`.  So this standard name would become:

`reference_pressure_in_atmosphere_layer_normalized_by_surface_reference_pressure`

Would that work for you @grantfirl ?  If so I'll go ahead and update both this standard name and the `reference_pressure` standard name.

nusbaume:
@mkavulich @grantfirl @cacraigucar I forgot to mention that I also added some changes to `meta_stdname_check.py` to ensure that the printed output maintains the same order everytime, which hopefully explains the new changes to that script.  Of course if you would like me to make a separate issue for it just let me know.  Thanks!

standard_names.xml
mkavulich: I think `tendency_of_dry_air_enthalpy_at_constant_pressure` is much better, despite how much the term "enthalpy" gives me undergrad thermodynamics flashbacks. We could also keep the standard name as "`heating_rate_of_dry_air`" and include the enthalpy descriptor as the long name.

It occurs to me that there is really no guidance at all for "long names" in the standards. Maybe a discussion topic for a future meeting?

cacraigucar (approved):

standard_names.xml
grantfirl: Yes, @nusbaume, the new standard name is much clearer. Thanks. @mkavulich Long names have always just functioned as documentation basically. The framework ignores them. They can be context-dependent, i.e. have differences between metadata files, if it makes sense to do so.

Metadata-standard-names.md
grantfirl: Yes, that is clearer. Thanks.

grantfirl:
@nusbaume I'll approve once you commit those couple of changes that we discussed. Please tag me with @ when you're done so that I get an email (that I hopefully see!).

standard_names.xml
nusbaume: @mkavulich @grantfirl The standard name has now been changed.  Thanks!

Metadata-standard-names.md
nusbaume: @grantfirl Great!  The standard names have now been changed.

grantfirl (approved):

Issue #26, opened 2021-11-05
Units for relative humidity


climbfuji:
Relative humidity in the ccpp-physics library / UFS / SCM takes all sorts of units, I found `1`, `frac`, `none`, and even `%`  so far. According to the standard name rules, the unit should be `1`. Please confirm and I will make the corresponding changes.

If those quantities that have `%` are indeed percentages, then do we need another standard name, or do we simply define `percent` as a valid unit (the symbol can cause all sorts of trouble) and register a unit conversion from percent to one and vice versa?

climbfuji:
> Relative humidity in the ccpp-physics library / UFS / SCM takes all sorts of units, I found `1`, `frac`, `none`, and even `%` so far. According to the standard name rules, the unit should be `1`. Please confirm and I will make the corresponding changes.
> 
> If those quantities that have `%` are indeed percentages, then do we need another standard name, or do we simply define `percent` as a valid unit (the symbol can cause all sorts of trouble) and register a unit conversion from percent to one and vice versa?

Update. Those quantities are not percentages, they have an upper bound of 1. Leaves the question if we should use 1 as per the rules (my preference) or any of the other units that are in the metadata.

gold2718:
`percent` is a valid unit in UDUNITS so I vote for that for true percentages. We can have it be equivalent to 1 (although it would be responsible to see if that ends up making sense or is just a lazy way to get things to work).

climbfuji:
Agreed. The quantities in the UFS/ccpp-physics library definitely range from 0 to 1 and are therefore not percentages. But we should add `percent` to the rules.

climbfuji:
@gold2718 EMC would prefer `fraction` as unit for relative humidity that is not a percentage, i.e. that ranges from zero to one, rather than `1`.  I am ok with fraction, and I think we can also add an "identify" unit conversion from `1` to `fraction` to allow both versions. What do you think? Also asking @cacraigucar @ligiabernardet @grantfirl for their opinion.

gold2718:
I agree that a value that varies between zero and one is a fraction and not a `percent` so relative humidity should be represented as a fraction (`1`). The unit, `percent` should be reserved for quantities that range from zero to one hundred.
`fraction` is not in UDUNITS and there is feeling among at least some of the maintainers that units should not include information about the physical quantity. I think we should stick with `1` rather than purposely ignore a standard.

junwang-noaa:
[...]

PR #62, opened 2024-03-20
Update Standard Name Rules table for dimensionless units, change variables improperly using "1" to "fraction"

mkavulich:
Related to Issue #26 and discussion therein, this PR updates the Standard Name Rules table for dimensionless units, removing some ambiguity in wording, separating out "frac" from "percent" entries (since fractions and percentages are different things, and can theoretically be converted) and clarifying when "1" should be used. In addition, several standard names that reference fractions are updated to appropriately use "frac" rather than "1".


nusbaume (commented):
Thanks for tackling this @mkavulich!  Would you be willing to add a couple extra unit changes to this PR?  Specifically the following:

1. `cellular_automata_seed_probability` currently has units of `fraction`.  Might be worth changing that to `frac`.
2.   `chemical_tracer_scavenging_fractions`  should probably also have units of `frac`.

Otherwise I think everything else looks good, at least to me.

climbfuji:
Why can we not spell out fraction? We also spell out percent. The 8-character limit in F77 is history!

mkavulich:
@climbfuji I'm just going off the existing conventions; all the schemes in ccpp-physics currently use `frac`. But if we're okay with committing to that change then I'm all for using `fraction`.

nusbaume:
I'm personally fine with either `frac` or `fraction`, just as long as we aren't using both of them together.  I also agree that `fraction` is certainly less ambiguous (but would require more effort on this particular PR).

gold2718 (approved):
Looks okay for me

climbfuji:
If it's not too much work, can you change all `frac` to `fraction` please?

cacraigucar (approved):

mkavulich:
@nusbaume @climbfuji I have made the suggested changes

climbfuji (approved):
Thanks for considering my comments!

nusbaume (approved):
Looks great to me now, thanks!

PR #71, opened 2024-09-09
Add miscellaneous atmosphere variables

svahl991:
This adds some new variable names, used in the JEDI system, that are mostly usual variants of existing CCPP variable names. They concern winds, temperature, and relative humidity.

~~Perhaps the newest concept introduced here is `surface_wind_from_direction`. I did not find any other existing CCPP variable names with direction expressed in degrees like this one.~~ removed from this PR

### DIFF
--- a/Metadata-standard-names.md
+++ b/Metadata-standard-names.md
@@ -129,6 +129,10 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
     * `real(kind=kind_phys)`: units = m s-1
 * `northward_wind`: Wind vector component, positive when directed northward
     * `real(kind=kind_phys)`: units = m s-1
+* `eastward_wind_at_10m`: Wind vector component at 10m, positive when directed eastward
+    * `real(kind=kind_phys)`: units = m s-1
+* `northward_wind_at_10m`: Wind vector component at 10m, positive when directed northward
+    * `real(kind=kind_phys)`: units = m s-1
 * `surface_wind_speed`: Scalar wind speed at surface
     * `real(kind=kind_phys)`: units = m s-1
 * `surface_wind_from_direction`: Direction, from north, of wind speed at surface
--- a/Metadata-standard-names.md
+++ b/Metadata-standard-names.md
@@ -129,18 +129,10 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
     * `real(kind=kind_phys)`: units = m s-1
 * `northward_wind`: Wind vector component, positive when directed northward
     * `real(kind=kind_phys)`: units = m s-1
-* `surface_eastward_wind`: Surface wind vector component, positive when directed eastward
-    * `real(kind=kind_phys)`: units = m s-1
-* `surface_northward_wind`: Surface wind vector component, positive when directed northward
-    * `real(kind=kind_phys)`: units = m s-1
 * `eastward_wind_at_10m`: Wind vector component at 10m, positive when directed eastward
     * `real(kind=kind_phys)`: units = m s-1
 * `northward_wind_at_10m`: Wind vector component at 10m, positive when directed northward
     * `real(kind=kind_phys)`: units = m s-1
-* `surface_wind_speed`: Scalar wind speed at surface
-    * `real(kind=kind_phys)`: units = m s-1
-* `surface_wind_from_direction`: Direction, from north, of wind speed at surface
-    * `real(kind=kind_phys)`: units = degrees
 * `dry_static_energy`: Dry static energy Content of Atmosphere Layer
     * `real(kind=kind_phys)`: units = J kg-1
 * `do_lagrangian_vertical_coordinate`: flag indicating if vertical coordinate is lagrangian
--- a/Metadata-standard-names.md
+++ b/Metadata-standard-names.md
@@ -253,6 +253,8 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
     * `real(kind=kind_phys)`: units = K
 * `virtual_temperature`: virtual temperature
     * `real(kind=kind_phys)`: units = K
+* `virtual_potential_temperature`: virtual potential temperature
+    * `real(kind=kind_phys)`: units = K
 * `composition_dependent_gas_constant_of_dry_air`: Composition dependent gas constant of dry air
     * `real(kind=kind_phys)`: units = J kg-1 K-1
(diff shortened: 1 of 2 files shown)


svahl991:
Tagging @BenjaminTJohnson since some of these variables are used with CRTM.

Metadata-standard-names.md
```diff
@@ -129,6 +129,18 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
     * `real(kind=kind_phys)`: units = m s-1
 * `northward_wind`: Wind vector component, positive when directed northward
     * `real(kind=kind_phys)`: units = m s-1
+* `surface_eastward_wind`: Surface wind vector component, positive when directed eastward
```
climbfuji, line 132: Isn't the convention to use `eastward_wind_at_surface`? I see you are adding `northward_wind_at_10m` explicitly below. I need to remind myself if the intention here is to not specify the suffix, because the rules say that the suffices listed elsewhere in the repo can be/are added to the root of the standard name?

svahl991: I was a little unsure about this, but ultimately decided to go with `surface_eastward_wind` because, less than 20 lines above this one (lines 114, 116, and 118 in this file) are `surface_air_pressure`, `surface_pressure_of_dry_air`, and `surface_geopotential`. So `eastward_wind_at_surface` seemed inconsistent.

mkavulich: It appears that in the CF conventions the standard phraseology is "surface_xxx" for variables at the surface. There are also existing examples `surface_northward_sea_water_velocity` and `surface_eastward_sea_water_velocity` (though that's not a perfect analogy since this is specifically the ocean surface). This unfortunately seems to be inconsistent with the phraseology used for every other variable at a specific level, which does follow the convention of `xxx_at_yyy` for a specific level `yyy`. So while it does seem awkward, I think these new names are in line with the existing conventions.

@svahl991 I do have a question though, can you clarify the difference between the "surface" variables and the "10m" variables? In the contexts I'm aware of, the "surface" wind field *is* the 10 m wind field. Is this some other level separate from the 10m level?

BenjaminTJohnson:
@mkavulich  My interpretation is that surface variables *try to be at the surface*, not 2m or 10m above the surface if something closer is available.      We do frequently use 2m or sometimes 10m variables to *replace* surface variables, but the strict definition of "surface" is the value exactly at the boundary (e.g., the temperature of the land, ocean, ice, snow, etc.).   Having said that,  "surface wind speed" is subject to all sorts of qualifications of what the actual physical interpretation is, but my understanding is that "surface wind speed" should be: "the available wind speed information that is closest to the surface".  If one only has 2m and 10m wind speed from the model, then you should probably use the 2m as the surface wind speed.  

If we want to get into semantics, and maybe we do, then we may wish to differentiate "near surface" from "surface" and "surface on the atmosphere side" vs. "surface on the surface side".     If you think of the surface as a boundary: one can approach that boundary "below" (e.g., from the ocean depths toward the ocean surface) vs. approaching it from "above" (e.g., from the atmosphere down to the ocean surface).      One clear example is temperature profile in the upper ocean, vs. temperature profile in the lower atmosphere.

mikecooke77:
For RTTOV the explanation @BenjaminTJohnson has given is also true.  The surface T for instance is used to calculate the near surface layer contribution to the radiance. The model generally produces 2m values apart from winds which are normally 10m.

As discussed previously RTTOV also has a skin temperature which can vary per channel and represents the temperature of the surface.

svahl991:
Given the responses above , I think it's been established that the "surface" variables, at least in principle, represent a different concept than the "2m" and "10m" variables, and so are worthy of their own names being added to CCPP. And having `surface_*` at the front of the variable names instead of `*_at_surface` as a suffix, is also the consistent way to go. So I don't think any required changes to the contents of this PR have been identified.

climbfuji:
We had a discussion between the CCPP standard names maintainers and the suggestion was to avoid the inconsistency that variables at the "surface" (whatever that is) have `surface_xyz` while every other level (layer, interface, ...) is added as a suffix. See #72 for a draft PR

gold2718 (approved):
LGTM
I do see that this will need to be revisited if #72 is accepted as is.

svahl991:
So, to make sure I understand, rather than change this PR, the plan is to accept it as it is for now, and then if/when #72 is accepted, all the variables that currently start with `surface_*` will be changed to `*_at_surface`?

On a related note, I have been wondering about the status of the [issue regarding the of tagging the naming standards](https://github.com/ESCOMP/CCPPStandardNames/issues/68) since they seem to be pretty fluid.

climbfuji:
As far as I understand, you hardcode the CCPP standard names into the actual C++ code. Therefore, you need to decide whether you want to have this merged and then change your standard names later when/if #72 is merged, or you want to do this all at once. (I don't have a preference, but if you want to merge this PR first then I will approve with a note that those names will liekly be changed alongside others.)

svahl991:
> As far as I understand, you hardcode the CCPP standard names into the actual C++ code. Therefore, you need to decide whether you want to have this merged and then change your standard names later when/if #72 is merged, or you want to do this all at once. (I don't have a preference, but if you want to merge this PR first then I will approve with a note that those names will liekly be changed alongside others.)

Given that we _do_ hardcode CCPP names into JEDI C++ code, and that it takes non-trivial coordinated effort to change them, and that it is only going to become more difficult to change them as JEDI becomes operational in more organizations, I would prefer do this all at once if #72 is going to be merged. We will be having a code sprint the week of Oct. 7 where we will updating many variable names in JEDI. So if it is highly probable that #72 will go through, we will probably want to switch to the new names during that sprint. If it is possible for us to know the fate of #72 before our sprint (even if it is not fully completed), that would be helpful.

Tagging @shlyaeva

svahl991:
Given the ongoing discussion regarding #72, I have now removed all variables with the `surface_*` prefix from this PR. I will submit another PR for those variable names when #72 is sorted out, hopefully soon. But I didn't want that issue to hold up progress on getting the other names in this PR added.

Metadata-standard-names.md
svahl991: All `surface_*` variables have now been removed from this PR, pending the resolution of #72.

climbfuji:
Can we have one more approval before we merge, please?

ss421 (approved):
Changes look good to me. Thankyou @svahl991

PR #73, opened 2024-09-13
Add four wind CF variables

svahl991:
This adds four variables related to wind fluid dynamics. The names are already CF names that have not yet been added to the CCPP standard.

### DIFF
--- a/standard_names.xml
+++ b/standard_names.xml
@@ -351,7 +351,7 @@
                    long_name="The curl of the vector wind field">
       <type kind="kind_phys" units="s-1">real</type>
     </standard_name>
-    <standard_name name="horizontal_divergence_of_wind"
+    <standard_name name="divergence_of_wind"
                    long_name="The divergence of the vector wind field">
       <type kind="kind_phys" units="s-1">real</type>
     </standard_name>
--- a/standard_names.xml
+++ b/standard_names.xml
@@ -352,7 +352,7 @@
       <type kind="kind_phys" units="s-1">real</type>
     </standard_name>
     <standard_name name="horizontal_divergence_of_wind"
-                   long_name="The divergence of the vector wind field>
+                   long_name="The divergence of the vector wind field">
       <type kind="kind_phys" units="s-1">real</type>
     </standard_name>
     <standard_name name="surface_upward_heat_flux_in_air">
--- a/standard_names.xml
+++ b/standard_names.xml
@@ -339,6 +339,22 @@
                    long_name="Total change in northward wind from a physics suite">
       <type kind="kind_phys" units="m s-2">real</type>
     </standard_name>
+    <standard_name name="atmosphere_horizontal_streamfunction"
+                   long_name="Scalar function describing the stream lines of the wind">
+      <type kind="kind_phys" units="m2 s-1">real</type>
+    </standard_name>
+    <standard_name name="atmosphere_horizontal_velocity_potential"
+                   long_name="Scalar potential of the wind">
+      <type kind="kind_phys" units="m2 s-1">real</type>
+    </standard_name>
+    <standard_name name="atmosphere_upward_absolute_vorticity"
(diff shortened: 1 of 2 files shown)


svahl991:
Tagging @ncrossette

climbfuji (approved):

standard_names.xml
```diff
@@ -339,6 +339,22 @@
...
+    </standard_name>
+    <standard_name name="atmosphere_horizontal_velocity_potential"
+                   long_name="Scalar potential of the wind">
+      <type kind="kind_phys" units="m2 s-1">real</type>
+    </standard_name>
+    <standard_name name="atmosphere_upward_absolute_vorticity"
+                   long_name="The curl of the vector wind field">
+      <type kind="kind_phys" units="s-1">real</type>
+    </standard_name>
+    <standard_name name="horizontal_divergence_of_wind"
```
gold2718, line 354: Isn't the CF name `divergence_of_wind` because CF considers "wind" to be the horizontal component?

gold2718 (changes requested):
I'm confused about one of the names.

standard_names.xml
svahl991: The note attached to `divergence of wind` (visible if you click on the name) in the [CF website](https://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html) says:

> "[horizontal_]divergence_of_X" means [horizontal] divergence of a vector X; if X does not have a vertical component then "horizontal" should be omitted. Wind is defined as a two-dimensional (horizontal) air velocity vector, with no vertical component. (Vertical motion in the atmosphere has the standard name upward_air_velocity.)

So now that I read that note more carefully, I see you are correct. Thanks for catching this. I think I misread the note originally. I will make this change.

Tagging @MarekWlasak to make sure he is OK with this change, since he suggested the name with `horizontal_` to me originally.

svahl991: PR has now been updated to remove `horizontal_` from `divergence_of_wind`.

standard_names.xml
```diff
@@ -339,6 +339,22 @@
...
+    <standard_name name="atmosphere_horizontal_velocity_potential"
+                   long_name="Scalar potential of the wind">
+      <type kind="kind_phys" units="m2 s-1">real</type>
+    </standard_name>
+    <standard_name name="atmosphere_upward_absolute_vorticity"
+                   long_name="The curl of the vector wind field">
+      <type kind="kind_phys" units="s-1">real</type>
+    </standard_name>
+    <standard_name name="divergence_of_wind"
+                   long_name="The divergence of the vector wind field">
```
gold2718, line 355: One option is to add horizontal in the long name.

gold2718 (approved):
Looks good now, thanks!

standard_names.xml
svahl991: > .... but wind does have 3 components (u,v w), so we do need the `horizontal_` bit otherwise we are talking about the 3 dimensional divergence!

I believe what the CF standard is saying in their comment is that wind is defined to be a 2-D variable, with the vertical component broken out as a separate variable (`upward_air_velocity`), and that's why the `horizontal_` prefix is not required in the `divergence_of_wind` name.

standard_names.xml
svahl991: I have now updated the long name to make it more descriptive and to specifically mention that wind is 2-D.

standard_names.xml
MarekWlasak: Ok. Just checked the CF standard.  The `wind` tag is indeed defined as implicitly `horizontal`, so no need for the prefix.  I am bit surprised by this though, as there is no tag for a 3D wind!

MarekWlasak (approved):

MarekWlasak:
My concerns and suggestions are in https://github.com/ESCOMP/CCPPStandardNames/issues/77

PR #74, opened 2024-09-19
Fix CODEOWNERS so that not only ss421 gets notified/must approve

climbfuji:
Update CODEOWNERS. See in the file itself: "Order is important. The last matching pattern has the most precedence."

See right hand side for a demonstration that the current file is wrong.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -3,19 +3,7 @@
 
 # These owners will be the default owners for everything in the repo.
 
-*       @cacraigucar      # NCAR/CGD
-*       @climbfuji        # NRL
-*       @dustinswales     # NOAA/GSL
-*       @gold2718         # Norway MET
-*       @grantfirl        # NOAA/GSL
-*       @mattldawson      # NCAR/ACOM
-*       @mkavulich        # DTC
-*       @mwaxmonsky       # NCAR/CGD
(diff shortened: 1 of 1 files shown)


climbfuji:
@cacraigucar @climbfuji @dustinswales @gold2718 @grantfirl @mattldawson @mkavulich @mwaxmonsky @nusbaume @peverwhee @svahl991 @MayeulDestouches - tagging you here instead of adding you as reviewers manually so that you can see the problem yourself.

ss421 (approved):
Thanks. Happy with the change. I think we should remove @MayeulDestouches from the list (or update) because he is no-longer at the UKMO. Perhaps that could be done in a seperate change.

climbfuji:
> Thanks. Happy with the change. I think we should remove @MayeulDestouches from the list (or update) because he is no-longer at the UKMO. Perhaps that could be done in a separate change.

Good point - do you want to nominate another UKMO representative in addition to yourself?

ss421:
> > Thanks. Happy with the change. I think we should remove @MayeulDestouches from the list (or update) because he is no-longer at the UKMO. Perhaps that could be done in a separate change.
> 
> Good point - do you want to nominate another UKMO representative in addition to yourself?

Yes that would be good. Ill see if I can get someone else that could be added. Thankyou.

PR #75, opened 2024-09-20
New wind variables at surface

svahl991:
This adds two pairs of wind variables for winds at the surface layer.

Note that the names here assume that the naming rule changes in #72 will be merged.

### DIFF
--- a/Metadata-standard-names.md
+++ b/Metadata-standard-names.md
@@ -133,13 +133,13 @@ Note that appending '_on_previous_timestep' to standard_names in this section yi
     * `real(kind=kind_phys)`: units = m s-1
 * `northward_wind_at_10m`: Wind vector component at 10m, positive when directed northward
     * `real(kind=kind_phys)`: units = m s-1
-* `eastward_wind_at_surface`: Wind vector component at surface, positive when directed eastward
+* `eastward_wind_at_surface`: Wind vector component closest to surface, positive when directed eastward
     * `real(kind=kind_phys)`: units = m s-1
-* `northward_wind_at_surface`: Wind vector component at surface, positive when directed northward
+* `northward_wind_at_surface`: Wind vector component closest to surface, positive when directed northward
     * `real(kind=kind_phys)`: units = m s-1
-* `wind_speed_at_surface`: Scalar wind speed at surface
+* `wind_speed_at_surface`: Scalar wind speed closest to surface
     * `real(kind=kind_phys)`: units = m s-1
-* `wind_from_direction_at_surface`: Direction, from north, of wind speed at surface
+* `wind_from_direction_at_surface`: Direction, from north, of wind speed closest to surface
     * `real(kind=kind_phys)`: units = degrees
 * `dry_static_energy`: Dry static energy Content of Atmosphere Layer
     * `real(kind=kind_phys)`: units = J kg-1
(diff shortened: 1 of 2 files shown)


svahl991:
Tagging @shlyaeva

gold2718 (approved):
LGTM

nusbaume (commented):
I am probably making a mountain out of a molehill, but what is the "surface" that these standard names are referring to?  For example my understanding is that when meteorologists report surface winds they are usually reporting winds at ten meters above the surface, which we already have a suffix for (`at_10m`).  

Alternatively my (extremely limited) understanding of boundary layer turbulence assumes that the wind speeds at the actual surface are exactly zero relative to the surface itself, so any local non-zero wind speed must be defined at some height above the actual surface.

Given this, if this is a quantity defined on model levels, would something like `at_surface_adjacent_layer` or `at_bottom_interface` be a better suffix?  

Anyways, I certainly won't hold up this PR for this as it sounds like it is needed soon, but I am just worried that we are potentially adding some ambiguity here.  Of course I'm happy to hear if people disagree.  Thanks!

climbfuji (approved):

ss421:
> I am probably making a mountain out of a molehill, but what is the "surface" that these standard names are referring to? For example my understanding is that when meteorologists report surface winds they are usually reporting winds at ten meters above the surface, which we already have a suffix for (`at_10m`).
> 
> Alternatively my (extremely limited) understanding of boundary layer turbulence assumes that the wind speeds at the actual surface are exactly zero relative to the surface itself, so any local non-zero wind speed must be defined at some height above the actual surface.
> 
> Given this, if this is a quantity defined on model levels, would something like `at_surface_adjacent_layer` or `at_bottom_interface` be a better suffix?
> 
> Anyways, I certainly won't hold up this PR for this as it sounds like it is needed soon, but I am just worried that we are potentially adding some ambiguity here. Of course I'm happy to hear if people disagree. Thanks!

I agree with this comment but I do not have anything useful to add as I'm not sure what `at_surface` means in particular since we have the `at_10m` suffix. I can see in JEDI that we have an "at surface" wind variable name. Would be great if the detail related to that could be added either in the name as suggested or in the description of the name. I appreciate that may not be possible and I do not want to hold this up so am approving in this basis.

ss421 (approved):

svahl991:
> I am probably making a mountain out of a molehill, but what is the "surface" that these standard names are referring to? 

The names in this PR were previously part of #71 (but were removed from that PR for other reasons), and this same discussion came up. See comments from @BenjaminTJohnson and @mikecooke77 on that PR.
https://github.com/ESCOMP/CCPPStandardNames/pull/71#issuecomment-2346810003

Also tagging @shlyaeva

nusbaume:
@svahl991 Thanks for pointing me to the extra discussion!  I think that all makes sense to me.  Just to make sure it is as clear as possible, could we possibly update the long names to make it obvious that these wind fields are whatever wind variable is closest to the surface?  So using `eastward_wind_at_surface` as an example, could the long name be:

` Wind vector component closest to surface, positive when directed eastward`

svahl991:
> So using eastward_wind_at_surface as an example, could the long name be:
 Wind vector component closest to surface, positive when directed eastward

I've now updated the long names in this PR with this clarification, as requested.

nusbaume (approved):
Thanks for humoring me!  Everything looks good on my end now.

climbfuji:
Thanks @svahl991. Let's wait for @mkavulich to review and merge?

mikecooke77 (approved):
Thanks for adding

climbfuji:
We are unfortunately still waiting for approval from the NOAA side. I pinged @mkavulich and @dustinswales in the other PR #76 a few days ago.

dustinswales (approved):
@climbfuji Sorry for the delay.
No objections from me.

PR #76, opened 2024-09-24
Add/modify variable names related to surface height/pressure

svahl991:
This adds two names relating to height at surface, and modifies `surface_air_pressure` for the naming rule change coming in #72.

(I realize that there is more renaming of existing variables to be done as a product of #72, so I can remove `surface_air_pressure` from this PR if people feel it is better to do that rename in a group with the others. It's just that this name is of particular interest to the fast-approaching JEDI renaming code sprint, so I wanted to be sure to get it in.)

### DIFF
--- a/standard_names.xml
+++ b/standard_names.xml
@@ -243,6 +243,10 @@
                    long_name="geopotential height w.r.t. sea level">
       <type kind="kind_phys" units="m">real</type>
     </standard_name>
+    <standard_name name="geopotential_height_at_surface"
+                   long_name="geopotential height at local surface">
+      <type kind="kind_phys" units="m">real</type>
+    </standard_name>
     <standard_name name="geopotential_height_wrt_surface"
                    long_name="geopotential height w.r.t. local surface">
       <type kind="kind_phys" units="m">real</type>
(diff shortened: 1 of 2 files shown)


gold2718 (approved):
Looks good!

climbfuji (approved):

standard_names.xml
```diff
@@ -243,6 +243,10 @@
                    long_name="geopotential height w.r.t. sea level">
       <type kind="kind_phys" units="m">real</type>
     </standard_name>
+    <standard_name name="geopotential_height_at_surface"
+                   long_name="geopotential height at local surface">
```
nusbaume, line 247: We might need to specify what this field is in relation to, for example:

```suggestion
                   long_name="geopotential height at local surface w.r.t. sea level">
```

nusbaume (changes requested):
Looks good!  Just had one request.

standard_names.xml
svahl991: I've now added `w.r.t. sea level`. Also capitalized `Geopotential` in `long_name` since that seems to be standard.

cacraigucar (approved):

ss421 (approved):

climbfuji:
@mkavulich @dustinswales this is ready for your final review and merging - thanks!

dustinswales (approved):
I see no issue with these changes/additions.

PR #80, opened 2024-10-24
Update StandardNamesRules.rst: add _due_to_planetary_boundary_layer

climbfuji:
Add process suffix `_due_to_planetary_boundary_layer` as proposed new suffix in `StandardNamesRules.rst`.

This isn't exactly describing a process (i.e. is different than `due_to_convection`), but people use this or the abbreviation `due_to_PBL` a lot. Should it be even longer, i.e. `_due_to_planetary_boundary_layer_processes` ?

### DIFF
--- a/StandardNamesRules.rst
+++ b/StandardNamesRules.rst
@@ -121,8 +121,9 @@ CCPP Standard Name Rules
    already exists and cannot be reused.
 
 #. Spell out acronyms unless they are obvious to a vast majority of
-   scientists/developers who may come across them. A
-   list of currently-used aliases is below.
+   scientists/developers who may come across them. A list of currently-used
+   aliases is below. Whenever such an alias exist, use the alias in the
+   standard name and the full term in the long name.
 
 #. For control-oriented variables, if the variable is a Fortran logical,
    use flag_for ``_X``. If it is any other data type, use control_for ``_X``. All flags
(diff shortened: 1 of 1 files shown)


nusbaume:
I think my two cents here are that I would prefer the longer `_due_to_planetary_boundary_layer_processes`, or maybe something more specific like `_due_to_planetary_boundary_layer_turbulent_mixing` (although I certainly realize it might be hard to disentangle every process if it is coming from a PBL physics scheme).

The reason I don't think I would want just `_due_to_planetary_boundary_layer` is because the PBL is also a location, and I could easily see variables like `X_in_planetary_boundary_layer`, which I feel could cause some confusion even with the different prepositions.

Of course I am happy to hear other people's thoughts as well, especially if they disagree.  Thanks!

gold2718:
Is `_due_to_PBL_processes` a decent compromise between length and clarity? I don't see any guidance in the CF conventions.

climbfuji:
> Is `_due_to_PBL_processes` a decent compromise between length and clarity? I don't see any guidance in the CF conventions.

We do spell out GWD as gravity_wave_drag, thus using PBL here would be inconsistent.

climbfuji:
At the 12/12/2024 CCPP meeting, we discussed using abbreviations like GWD and PBL **instead** of the full names in the standard names. The rules need to clarify that the short names must be used instead of the long names whenever such an abbreviation is defined, and that we will spell out the full name in the `long_name` instead. I will update this PR and request reviews one more time.

grantfirl:
There is an abbreviaton/acronym section in the rules (https://github.com/ESCOMP/CCPPStandardNames/blob/main/StandardNamesRules.rst). We can shorten names as needed and just add the acronym to the accepted list. PBL and GWD are good examples that are relatively well known in atmospheric sciences, and we can use the rules page to define them for those who are unfamiliar.

climbfuji:
I updated the PR as discussed this morning (hope I got it right). Please (re-)review. Thanks.

mkavulich (approved):

nusbaume (approved):

PR #84, opened 2024-11-08
Add/modify skin temperature at surface variables.

svahl991:
This PR modifies the existing `surface_skin_temperature` name to match the new naming rule introduced in #72. It also adds four new skin temperature variable variants. All these variables were introduced into JEDI during the October 2024 variable naming sprint.

Note a small JEDI/CCPP naming discrepancy discussed in https://github.com/ESCOMP/CCPPStandardNames/issues/83. I attempted to smooth over this difference in the long name.

### DIFF
--- a/standard_names.xml
+++ b/standard_names.xml
@@ -2826,7 +2826,8 @@
       <standard_name name="surface_roughness_length_over_water">
          <type kind="kind_phys" units="cm">real</type>
       </standard_name>
-      <standard_name name="skin_temperature_at_surface">
+      <standard_name name="skin_temperature_at_surface"
+                     long_name="Skin temperature at surface">
          <type kind="kind_phys" units="K">real</type>
       </standard_name>
       <standard_name name="skin_temperature_at_surface_over_ice"
--- a/standard_names.xml
+++ b/standard_names.xml
@@ -2829,16 +2829,20 @@
       <standard_name name="skin_temperature_at_surface">
          <type kind="kind_phys" units="K">real</type>
       </standard_name>
-      <standard_name name="skin_temperature_at_surface_over_ice">
+      <standard_name name="skin_temperature_at_surface_over_ice"
+                     long_name="Skin temperature at surface over (or where) ice">
          <type kind="kind_phys" units="K">real</type>
       </standard_name>
-      <standard_name name="skin_temperature_at_surface_over_land">
+      <standard_name name="skin_temperature_at_surface_over_land"
+                     long_name="Skin temperature at surface over (or where) land">
          <type kind="kind_phys" units="K">real</type>
       </standard_name>
-      <standard_name name="skin_temperature_at_surface_over_sea">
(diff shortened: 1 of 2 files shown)


svahl991:
Tagging @travissluka @rtodling @mikecooke77 @fmahebert

climbfuji (commented):
I believe we use `over_ocean` instead of `over_sea` (and we use `over_lake`; in case we don't need to distinguish, we use `over_water`). I checked the standard name dictionary, there are two entries with `over_ocean` and none with `over_sea`. In the NCAR ccpp-physics repo, there are six entries with `over_ocean`, but none with `over_sea`.

This can be changed if there is a good reason to use sea and not ocean - is there?

svahl991:
> This can be changed if there is a good reason to use sea and not ocean - is there?

I will wait a bit for the scientists to chime in. If I don't hear anything I'll change the name in this PR to `ocean` early next week.

standard_names.xml
```diff
@@ -2826,10 +2826,24 @@
...
          <type kind="kind_phys" units="cm">real</type>
       </standard_name>
-      <standard_name name="surface_skin_temperature">
+      <standard_name name="skin_temperature_at_surface"
+                     long_name="Skin temperature at surface">
          <type kind="kind_phys" units="K">real</type>
       </standard_name>
-      <standard_name name="surface_skin_temperature_over_land">
+      <standard_name name="skin_temperature_at_surface_over_ice"
+                     long_name="Skin temperature at surface over (or where) ice">
```
gold2718, line 2834: Is any distinction needed between ice over land and ice over ocean?

gold2718 (approved):
Looks okay
I have a question about the meaning (or generality) of the term ice but do not want it to hold up this PR.

standard_names.xml
svahl991: > Is any distinction needed between ice over land and ice over ocean?

 @travissluka we discussed this in-person and you had some thoughts?

travissluka: good point, it's probably best to keep `skin_temperature_at_surface_over_ice` for when we don't care if it is land ice or ocean ice, and have separate variables for land and ocean.
@svahl991 I don't think we do anything with land ice at the moment (in my world all ice is ocean ice) but  distinguishing between the two might be needed in the future

nusbaume (approved):
Approving with the assumption that `over_sea` will be changed to `over_ocean`.

mkavulich:
@svahl991 Just checking in on the status of this PR, assuming that you're okay with the change of `over_sea` --> `over_ocean` it looks like this PR can be merged after that change.

svahl991:
Thanks for the ping @mkavulich . This fell off my radar. I have now changed "sea" to "ocean" (after consulting with @travissluka ), so hopefully this is good to go now.

climbfuji (approved):

PR #86, opened 2024-12-11
add emissions names

jeromebarre:
This adds emissions names to the CCPP standard names

PR #87, opened 2024-12-13
Second rules update in v1 branch, update names, update `write_standard_name_table.py` to allow for subsections

mkavulich:
This is the second round of proposed rules changes based on our ongoing discussion to make both rules and names as consistent as possible. Major updates include:

### Standard Names
 - New "base_names" section, with subsections
 - Convert instances of "surface_X" to "X_at_surface"
 - Convert some instances of "air_X" or "X_of_air"
 - Explicitly state what mass mixing ratios are with respect to in long_name
 - Convert "mixing_ratio" to "water_vapor_mixing_ratio_wrt_moist_air"
 - Convert "surface_albedo" to simply "albedo", "surface_roughness_length" to simply "roughness_length"

### Rules
 - Update construction template to include [non-instant time] and [non-current time], swap [at level] and [in medium]
 - More detailed description of transformations and how they can work on multiple base names
 - Add a few more transformations

### `write_standard_name_table.py`
 - Updated to allow sub-sections in the standard names list

You can see these changes in the form of a Google doc for better visualization here: https://docs.google.com/document/d/19ysUCWDhv53W8fQbElW7opr_1Pm7ck95QRUyKM_qy4E/edit?tab=t.0 

Note: with this update, we have 347 standard names that are fully compliant with the rules we have set out. A big portion of the remainder are "flag"-type variables, indices, etc, which are not fully accounted for in the rules yet.

### DIFF
--- a/Metadata-standard-names.md
+++ b/Metadata-standard-names.md
@@ -1,7 +1,6 @@
 # CCPP Standard Name Library
 #### Table of Contents
 * [base_names](#base_names)
-* [base_standard_names](#base_standard_names)
 * [dimensions](#dimensions)
 * [constants](#constants)
 * [coordinates](#coordinates)
@@ -146,7 +145,7 @@ These are the base names for specific chemical species
 * `silicate`: Chemical species containing the silicate ion
 * `sulfate`: Chemical species containing the sulfate ion
 * `sulfur_dioxide`: so2
-## base_standard_names
+### base_standard_names
 These names are used as bases for other names, but may
 also be considered standard names on their own. See the
 full list of standard names for further details.
--- a/Metadata-standard-names.md
+++ b/Metadata-standard-names.md
@@ -41,8 +41,6 @@ more specific standard names.
     * `integer`: units = 1
 * `coefficient`: coefficient
     * `real`: units = 1
-* `coefficient`: coefficient
-    * `real`: units = 1
 * `data_mask`: data_mask
     * `real`: units = 1
 * `density`: density
--- a/Metadata-standard-names.md
+++ b/Metadata-standard-names.md
@@ -1,5 +1,6 @@
 # CCPP Standard Name Library
 #### Table of Contents
+* [base_names](#base_names)
 * [dimensions](#dimensions)
 * [constants](#constants)
 * [coordinates](#coordinates)
@@ -22,6 +23,100 @@
 * [GFS_typedefs_GFS_grid_type](#gfs_typedefs_gfs_grid_type)
 * [GFS_typedefs_GFS_stateout_type](#gfs_typedefs_gfs_stateout_type)
 
+## base_names
+Base names are the 'elemental' quantities from which
+the more complex standard names are constructed.
+Base names can roughly be broken down into three categories:
+### generic_names
+The following names are too general to be chosen as
+standard names, but they can serve as base names for
+more specific standard names.
+* `amount`: amount
+    * `real`: units = kg m-2
+* `area`: area
+    * `real`: units = m2
+* `area_fraction`: area_fraction
+    * `real`: units = 1
+* `binary_mask`: binary_mask
+    * `integer`: units = 1
+* `coefficient`: coefficient
+    * `real`: units = 1
+* `coefficient`: coefficient
+    * `real`: units = 1
+* `data_mask`: data_mask
+    * `real`: units = 1
+* `density`: density
+    * `real`: units = kg m-3
+* `energy`: energy
+    * `real`: units = J
+* `energy_content`: energy_content
+    * `real`: units = J m-2
+* `energy_density`: energy_density
+    * `real`: units = J m-3
+* `frequency`: frequency
+    * `real`: units = s-1
+* `heat_flux`: heat_flux
+    * `real`: units = W m-2
+* `heat_transport`: heat_transport
+    * `real`: units = W
+* `mass`: mass
+    * `real`: units = kg
+* `mass_flux`: mass_flux
+    * `real`: units = kg m-2 s-1
+* `mass_fraction`: mass_fraction
+    * `real`: units = 1
+* `mixing_ratio`: mixing_ratio
+    * `real`: units = kg kg-1
+* `mass_transport`: mass_transport
+    * `real`: units = kg s-1
+* `mole_fraction`: mole_fraction
+    * `real`: units = 1
+* `mole_flux`: mole_flux
+    * `real`: units = mol m-2 s-1
+* `momentum_flux`: momentum_flux
+    * `real`: units = Pa
+* `partial_pressure`: partial_pressure
+    * `real`: units = Pa
+* `period`: period
+    * `real`: units = s
+* `power`: power
+    * `real`: units = W
+* `pressure`: pressure
+    * `real`: units = Pa
+* `probability`: probability
+    * `real`: units = 1
+* `radiative_flux`: radiative_flux
+    * `real`: units = W m-2
+* `radius`: radius
+    * `real`: units = m
+* `specific_eddy_kinetic_energy`: specific_eddy_kinetic_energy
+    * `real`: units = m2 s-2
+* `speed`: speed
+    * `real`: units = m s-1
+* `stress`: stress
+    * `real`: units = Pa
+* `streamfunction`: streamfunction
+    * `real`: units = m2 s-1
+* `temperature`: temperature
+    * `real`: units = K
+* `thickness`: thickness
+    * `real`: units = m
+* `velocity`: velocity
+    * `real`: units = m s-1
+* `velocity_potential`: velocity_potential
+    * `real`: units = m2 s-1
+* `volume`: volume
+    * `real`: units = m3
+* `volume_flux`: volume_flux
+    * `real`: units = m s-1
+* `volume_fraction`: volume_fraction
+    * `real`: units = 1
+* `volume_mixing_ratio`: volume_mixing_ratio
+    * `real`: units = mol mol-1
+* `volume_transport`: volume_transport
+    * `real`: units = m3 s-1
+* `vorticity`: vorticity
+    * `real`: units = s-1
 ## dimensions
 Dimension standard names may come in sets of six related standard names for each dimension:
 ```
--- a/Metadata-standard-names.md
+++ b/Metadata-standard-names.md
@@ -154,7 +154,6 @@ full list of standard names for further details.
 * `air_pressure`: air_pressure
 * `air_pressure_thickness`: air_pressure_thickness
 * `air_temperature`: air_temperature
-* `air_temperature/water_vapor`: air_temperature/water_vapor
 * `albedo`: albedo
 * `atmosphere_heat_diffusivity`: atmosphere_heat_diffusivity
 * `cloud_area_fraction`: cloud_area_fraction
@@ -163,14 +162,14 @@ full list of standard names for further details.
 * `cloud_liquid_water`: cloud_liquid_water
 * `date`: date
 * `density`: density
-* `diffuse_nir_albedo`: diffuse_nir_albedo
-* `diffuse_nir_shortwave_flux`: diffuse_nir_shortwave_flux
+* `diffuse_nir_albedo`: diffuse near-infrared albedo
+* `diffuse_nir_shortwave_flux`: diffuse near-infrared shortwave flux
 * `diffuse_shortwave_albedo`: diffuse_shortwave_albedo
-* `diffuse_uv_and_vis_shortwave_flux`: diffuse_uv_and_vis_shortwave_flux
+* `diffuse_uv_and_vis_shortwave_flux`: diffuse ultraviolet and visible shortwave flux
 * `diffuse_visible_albedo`: diffuse_visible_albedo
-* `direct_nir_albedo`: direct_nir_albedo
-* `direct_nir_shortwave_flux`: direct_nir_shortwave_flux
-* `direct_uv_and_vis_shortwave_flux`: direct_uv_and_vis_shortwave_flux
+* `direct_nir_albedo`: direct near-infrared albedo
+* `direct_nir_shortwave_flux`: direct near-infrared shortwave flux
+* `direct_uv_and_vis_shortwave_flux`: direct ultraviolet and visible shortwave flux
 * `direct_visible_albedo`: direct_visible_albedo
 * `divergence`: divergence
 * `dry_air_density`: dry_air_density
@@ -214,14 +213,13 @@ full list of standard names for further details.
 * `temperature_flux`: temperature_flux
 * `time`: time
 * `total_energy`: total_energy
-* `total_water`: total_water
+* `total_water`: All water phases (solid, liquid, gas)
 * `tracer`: tracer
 * `tracers`: tracers
 * `turbulent_kinetic_energy`: turbulent_kinetic_energy
 * `velocity_potential`: velocity_potential
 * `virtual_potential_temperature`: virtual_potential_temperature
 * `virtual_temperature`: virtual_temperature
-* `water`: water
 * `water_vapor`: water_vapor
 * `wind`: wind
 * `wind_speed`: wind_speed
@@ -2067,25 +2065,25 @@ Variables related to the compute environment
     * `real(kind=kind_phys)`: units = 1
 * `convective_cloud_condensate_after_rainout`: Convective cloud condensate after rainout
     * `real(kind=kind_phys)`: units = kg kg-1
-* `cumulative_downwelling_diffuse_nir_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep`: Cumulative downwelling diffuse nir shortwave flux at surface for coupling multiplied by timestep
+* `cumulative_downwelling_diffuse_nir_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep`: cumulative downwelling diffuse near-infrared shortwave flux at the surface level for coupling multiplied by the duration of the timestep
     * `real(kind=kind_phys)`: units = J m-2
-* `cumulative_downwelling_diffuse_uv_and_vis_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep`: Cumulative downwelling diffuse uv and vis shortwave flux at surface for coupling multiplied by timestep
+* `cumulative_downwelling_diffuse_uv_and_vis_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep`: cumulative downwelling diffuse ultraviolet and visible shortwave flux at the surface level for coupling multiplied by the duration of the timestep
     * `real(kind=kind_phys)`: units = J m-2
-* `cumulative_downwelling_direct_nir_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep`: Cumulative downwelling direct nir shortwave flux at surface for coupling multiplied by timestep
+* `cumulative_downwelling_direct_nir_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep`: cumulative downwelling direct near-infrared shortwave flux at the surface level for coupling multiplied by the duration of the timestep
     * `real(kind=kind_phys)`: units = J m-2
-* `cumulative_downwelling_direct_uv_and_vis_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep`: Cumulative downwelling direct uv and vis shortwave flux at surface for coupling multiplied by timestep
+* `cumulative_downwelling_direct_uv_and_vis_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep`: cumulative downwelling direct ultraviolet and visible shortwave flux at the surface level for coupling multiplied by the duration of the timestep
     * `real(kind=kind_phys)`: units = J m-2
 * `cumulative_downwelling_longwave_flux_at_surface_for_coupling_multiplied_by_timestep`: Cumulative downwelling longwave flux at surface for coupling multiplied by timestep
     * `real(kind=kind_phys)`: units = J m-2
 * `cumulative_downwelling_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep`: Cumulative downwelling shortwave flux at surface for coupling multiplied by timestep
     * `real(kind=kind_phys)`: units = J m-2
-* `cumulative_net_downwelling_diffuse_nir_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep`: Cumulative net downwelling diffuse nir shortwave flux at surface for coupling multiplied by timestep
+* `cumulative_net_downwelling_diffuse_nir_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep`: cumulative net downwelling diffuse near-infrared shortwave flux at the surface level for coupling multiplied by the duration of the timestep
     * `real(kind=kind_phys)`: units = J m-2
-* `cumulative_net_downwelling_diffuse_uv_and_vis_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep`: Cumulative net downwelling diffuse uv and vis shortwave flux at surface for coupling multiplied by timestep
+* `cumulative_net_downwelling_diffuse_uv_and_vis_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep`: cumulative net downwelling diffuse ultraviolet and visible shortwave flux at the surface level for coupling multiplied by the duration of the timestep
     * `real(kind=kind_phys)`: units = J m-2
-* `cumulative_net_downwelling_direct_nir_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep`: Cumulative net downwelling direct nir shortwave flux at surface for coupling multiplied by timestep
+* `cumulative_net_downwelling_direct_nir_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep`: cumulative net downwelling direct near-infrared shortwave flux at the surface level for coupling multiplied by the duration of the timestep
     * `real(kind=kind_phys)`: units = J m-2
-* `cumulative_net_downwelling_direct_uv_and_vis_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep`: Cumulative net downwelling direct uv and vis shortwave flux at surface for coupling multiplied by timestep
+* `cumulative_net_downwelling_direct_uv_and_vis_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep`: cumulative net downwelling direct ultraviolet and visible shortwave flux at the surface level for coupling multiplied by the duration of the timestep
     * `real(kind=kind_phys)`: units = J m-2
 * `cumulative_net_downwelling_longwave_flux_at_surface_for_coupling_multiplied_by_timestep`: Cumulative net downwelling longwave flux at surface for coupling multiplied by timestep
     * `real(kind=kind_phys)`: units = J m-2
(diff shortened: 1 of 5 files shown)


standard_names.xml
```diff
@@ -1,4 +1,500 @@
...
+      </standard_name>
+      <standard_name name="binary_mask"
+                     long_name="binary_mask">
+        <type units="1">integer</type>
+      </standard_name>
+      <standard_name name="coefficient"
+                     long_name="coefficient">
+        <type units="1">real</type>
+      </standard_name>
+      <standard_name name="coefficient"
```
climbfuji, line 30: Duplicate of line 26-29

standard_names.xml
```diff
@@ -1,4 +1,500 @@
...
+                     long_name="Nitrogen dioxide">
+      </standard_name>
+      <standard_name name="no"
+                     long_name="Nitric oxide NO (Nitrogen oxide, Nitrogen monoxide)">
+      </standard_name>
+      <standard_name name="oxygen"
+                     long_name="Molecular oxygen, O₂">
+      </standard_name>
+      <standard_name name="ozone"
+                     long_name="Ozone, O₃">
```
climbfuji, line 249: Will these non-ASCII characters work? (also line 246)

standard_names.xml
```diff
@@ -2311,7 +2831,8 @@
       <standard_name name="upward_virtual_potential_temperature_flux">
          <type kind="kind_phys" units="K m s-1">real</type>
       </standard_name>
-      <standard_name name="surface_upward_specific_humidity_flux_for_mellor_yamada_janjic_surface_layer_scheme">
+      <standard_name name="upward_flux_of_water_vapor_mixing_ratio_wrt_moist_air_at_surface_for_mellor_yamada_janjic_surface_layer_scheme"
```
climbfuji, line 2834: We should probably define MYJ as the abbreviation to use in the standard name, and the full name in the long name? Like PBL and GWD in my open PR?

standard_names.xml
```diff
@@ -2365,31 +2890,35 @@
...
       </standard_name>
-      <standard_name name="surface_air_pressure_two_timesteps_back">
+      <standard_name name="air_pressure_at_surface_two_timesteps_back">
          <type kind="kind_phys" units="Pa">real</type>
       </standard_name>
       <standard_name name="control_for_surface_layer_evaporation">
          <type kind="kind_phys" units="1">real</type>
       </standard_name>
-      <standard_name name="surface_specific_humidity_for_myj_schemes">
+      <standard_name name="water_vapor_mixing_ratio_wrt_moist_air_at_surface_for_myj_schemes"
```
climbfuji, line 2913: Here, MYJ should be uppercase in the standard name, and spelled out in the long name

standard_names.xml
```diff
@@ -3074,70 +3636,74 @@
...
-      <standard_name name="surface_nir_albedo_direct_rad_for_coupling">
+      <standard_name name="direct_nir_albedo_for_coupling"
+                     long_name="surface albedo for direct near-infrared radiation for coupling">
          <type kind="kind_phys" units="fraction">real</type>
       </standard_name>
       <standard_name name="lwe_surface_snow_from_coupled_process">
          <type kind="kind_phys" units="m">real</type>
       </standard_name>
-      <standard_name name="surface_upward_latent_heat_flux_from_coupled_process">
+      <standard_name name="upward_latent_heat_flux_at_surface_from_coupled_process">
```
climbfuji, line 3671: What is the difference between `for_coupling` and `from_coupled_process` (not for vs from, but coupling vs coupled process)?

standard_names.xml
gold2718: Rule 17 says names are insensitive, does it matter here?
If so, maybe we should add something and uppercase the abbreviations in the [Acronyms, Abbreviations, and Aliases](https://github.com/ESCOMP/CCPPStandardNames/blob/main/StandardNamesRules.rst#acronyms-abbreviations-and-aliases) section? Or is it just for initials of names?

climbfuji: That (rule 17) is a good point I completely forgot about. I like consistency and would prefer either lowercase or uppercase. We can talk about it at one of the next meetings, if the majority thinks that case insensitive is fine, then I can live with that, too.

standard_names.xml
```diff
@@ -2924,52 +3484,52 @@
       <standard_name name="convective_cloud_condensate_after_rainout">
          <type kind="kind_phys" units="kg kg-1">real</type>
       </standard_name>
-      <standard_name name="cumulative_surface_downwelling_diffuse_nir_shortwave_flux_for_coupling_multiplied_by_timestep">
+      <standard_name name="cumulative_downwelling_diffuse_nir_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep">
          <type kind="kind_phys" units="J m-2">real</type>
       </standard_name>
-      <standard_name name="cumulative_surface_downwelling_diffuse_uv_and_vis_shortwave_flux_for_coupling_multiplied_by_timestep">
+      <standard_name name="cumulative_downwelling_diffuse_uv_and_vis_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep">
```
gold2718, line 3490: I think it would be helpful to spell out `uv_and_vis_shortwave` in a long name:
```suggestion
      <standard_name name="cumulative_downwelling_diffuse_uv_and_vis_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep"
      long_name="Cumulative downwelling diffuse ultraviolet_and_visible parts of the shortwave flux at surface for coupling multiplied by timestep">
```
This suggestion also applies to other uses of `uv_and_vis_shortwave`.

gold2718 (changes requested):
Great (and great big) piece of work!
A few requested changes and a lot of questions along with a few suggestions.
Also, should all long names begin with an capitalized word? Right now, there is a mix.

standard_names.xml
mkavulich: `long_name` should be able to handle any unicode characters since it is not used in any FORTRAN code context. Even if long names were eventually used in comments or even character variables, that would be fine according to the [FORTRAN 90 standard](https://wg5-fortran.org/N001-N1100/N692.pdf)*, so long as it's a character set supported by the OS. I think anything in UTF-32 should be allowed for long names, since Python can handle those characters and it's essentially universally supported among OS's. 

This does bring up the fact that we haven't defined a character set for the standard names; I'll add that to the to-do list. So long as we make anything that may possibly be used as a fortran variable/subroutine/other code strictly within the allowed Fortran character set, we should be good.

* The relevant section on page 18 is here:

>Additional characters may be representable in the processor, but may appear only in comments (3.3.1.1,
15 3.3.2.1), character constants (4.4.4), input/output records (9.1.1), and character string edit descriptors
16 (10.2.1).

standard_names.xml
mkavulich: I think this is a good idea, but I will add it to my next round of changes to avoid scope creep. There are probably more abbreviations to introduce beyond this, like MYNN.

standard_names.xml
mkavulich: This is a good point: In general I think the use of "coupling" and "coupled" in variable names is ambiguous and problematic. Probably needs a deeper dive after this PR.

standard_names.xml
```diff
@@ -2431,7 +2962,8 @@
       <standard_name name="weight_for_potential_temperature_at_top_of_viscous_sublayer">
          <type kind="kind_phys" units="1">real</type>
       </standard_name>
-      <standard_name name="weight_for_specific_humidity_at_top_of_viscous_sublayer">
+      <standard_name name="weight_for_water_vapor_mixing_ratio_wrt_moist_air_at_top_of_viscous_sublayer"
+                   long_name="Weight for specific humidity (water vapor mass mixing ratio with respect to moist air) at the top of the viscous sublayer">
```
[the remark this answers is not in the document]
mkavulich: It looks like "weight" is used to mean "weighting factor" in all current contexts. I'm not sure if referencing physical "weight" would ever be useful in standard names, but maybe it's worthy of discussion?

standard_names.xml
```diff
@@ -2589,16 +3125,20 @@
       <standard_name name="max_vegetation_area_fraction">
          <type kind="kind_phys" units="fraction">real</type>
       </standard_name>
-      <standard_name name="nir_albedo_strong_cosz">
+      <standard_name name="nir_albedo_strong_cosz"
```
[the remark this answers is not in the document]
mkavulich: Yes, I'll add it to the list for next time since I have collected a bunch of acronyms to add

standard_names.xml
```diff
@@ -2326,10 +2847,12 @@
...
          <type kind="kind_phys" units="m">real</type>
       </standard_name>
-      <standard_name name="specific_humidity_on_previous_timestep">
+      <standard_name name="water_vapor_mixing_ratio_wrt_moist_air_on_previous_timestep"
+                   long_name="Specific humidity (water vapor mass mixing ratio with respect to moist air) on previous timestep">
          <type kind="kind_phys" units="kg kg-1">real</type>
       </standard_name>
-      <standard_name name="tendency_of_specific_humidity_due_to_nonphysics">
+      <standard_name name="tendency_of_water_vapor_mixing_ratio_wrt_moist_air_due_to_nonphysics"
+                   long_name="Tendency of specific humidity (water vapor mass mixing ratio with respect to moist air) due to nonphysics">
```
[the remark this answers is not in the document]
mkavulich: It looks like this was one of the original standard names, so this was before my time. My impression is that this would refer to all non-physics processes, including the dycore and any other potential sources/sinks

standard_names.xml
```diff
@@ -1,4 +1,500 @@
...
+      <standard_name name="sulfur_dioxide"
+                     long_name="so2">
+      </standard_name>
+    </section>
+    <section name="base_standard_names"
+             comment="These names are used as bases for other names, but may\n
+                      also be considered standard names on their own. See the\n
+                      full list of standard names for further details.\n">
+      <standard_name name="absolute_vorticity"
+                     long_name="absolute_vorticity">
```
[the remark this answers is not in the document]
mkavulich: My intention is to include long names for all the variables in this section, but I hadn't gotten around to it for this round of commits. I can remove them for now if it's a big deal but I just wanted to note this is not intended to be the "final" version.

standard_names.xml
```diff
@@ -1,4 +1,500 @@
...
+           comment="Base names are the 'elemental' quantities from which\n
+                    the more complex standard names are constructed.\n
+                    Base names can roughly be broken down into three categories:\n">
+    <section name="generic_names"
+             comment="The following names are too general to be chosen as\n
+                      standard names, but they can serve as base names for\n
+                      more specific standard names.\n">
+      <standard_name name="amount"
+                     long_name="amount">
+        <type units="kg m-2">real</type>
```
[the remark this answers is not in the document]
mkavulich: These units were carried over as-is from a table that is in [https://github.com/ESCOMP/CCPPStandardNames/blob/release/v1/StandardNamesRules.rst#generic-names](StandardNamesRules.rst) (not yet removed). I do agree that if these base names are only intended to be components of other standard names then they do not necessarily need units, but on the other hand it is useful for many of these to specifically designate the units/dimensionality of the variable. So I could go either way on this, maybe calls for further discussion.

StandardNamesRules.rst
```diff
@@ -32,14 +32,14 @@ CCPP Standard Name Rules
...
-   [``transformation``] [``component``] base_name [*at* ``level``] [*in* ``medium``] [*due_to* ``process``] [*assuming* ``condition``]
+   [``transformation``] [``component``] [``non-instant time``] base_name [*in* ``medium``] [*at* ``level``] [*due_to* ``process``] [``non-current time``] [*assuming* ``condition``]
 
    This construction was originally based on rules set forth in the
    `CF guidelines <http://cfconventions.org/Data/cf-standard-names/docs/guidelines.html>`_),
    but have since evolved for better consistency and generality across a broader set of fields
    than was originally envisioned by the CF conventions. "``medium``" should be specified when
-   the variable in question is a substance or other quantity contained within some other medium
-   (e.g. for "mole_fraction_of_ozone_in_air", the base name is "ozone", while the medium is "air"). 
+   the variable in question is a substance or other quantity contained within some other the medium
```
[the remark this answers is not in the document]
mkavulich: Fixed

StandardNamesRules.rst
```diff
@@ -429,13 +446,13 @@ Special phrases
 +------------------------+-------------------------------------------------------------------------------------+
 |frozen_water            | ice                                                                                 |
 +------------------------+-------------------------------------------------------------------------------------+
-| longwave               | longwave radiation                                                                  |
+| longwave               | Longwave radiation. Defined as thermal emission of EM radiation from the planet.    |
```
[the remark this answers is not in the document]
mkavulich: Done

StandardNamesRules.rst
```diff
@@ -429,13 +446,13 @@ Special phrases
...
 +------------------------+-------------------------------------------------------------------------------------+
 | moisture               | water in all phases contained in soil                                               |
 +------------------------+-------------------------------------------------------------------------------------+
 | ocean                  | used instead of in_sea_water for quantities which are large-scale rather than local |
 +------------------------+-------------------------------------------------------------------------------------+
-| shortwave              | shortwave radiation                                                                 |
+| shortwave              | Shortwave radiation. Defined as EM emissions from the sun                           |
```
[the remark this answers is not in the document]
mkavulich: Done

standard_names.xml
mkavulich: Removed

standard_names.xml
```diff
@@ -1,4 +1,500 @@
...
+      </standard_name>
+      <standard_name name="air_pressure_thickness"
+                     long_name="air_pressure_thickness">
+      </standard_name>
+      <standard_name name="air_temperature"
+                     long_name="air_temperature">
+      </standard_name>
+      <standard_name name="air_temperature/water_vapor"
+                     long_name="air_temperature/water_vapor">
+      </standard_name>
```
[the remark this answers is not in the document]
mkavulich: Fixed, not sure how that snuck in there

standard_names.xml
mkavulich: Done

standard_names.xml
```diff
@@ -2924,52 +3484,52 @@
...
-      <standard_name name="cumulative_surface_downwelling_diffuse_nir_shortwave_flux_for_coupling_multiplied_by_timestep">
+      <standard_name name="cumulative_downwelling_diffuse_nir_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep">
          <type kind="kind_phys" units="J m-2">real</type>
       </standard_name>
-      <standard_name name="cumulative_surface_downwelling_diffuse_uv_and_vis_shortwave_flux_for_coupling_multiplied_by_timestep">
+      <standard_name name="cumulative_downwelling_diffuse_uv_and_vis_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep">
          <type kind="kind_phys" units="J m-2">real</type>
       </standard_name>
-      <standard_name name="cumulative_surface_downwelling_direct_nir_shortwave_flux_for_coupling_multiplied_by_timestep">
+      <standard_name name="cumulative_downwelling_direct_nir_shortwave_flux_at_surface_for_coupling_multiplied_by_timestep">
```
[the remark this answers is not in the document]
mkavulich: Done

mkavulich (commented):
@gold2718 @climbfuji Thanks for your thorough reviews. I've addressed most of your comments/suggestions, and those that I haven't I replied with a follow-up comment or question. Let me know if anything else needs resolving.

standard_names.xml
```diff
@@ -3178,19 +3758,23 @@
...
       <standard_name name="mass_number_concentration_of_cloud_liquid_water_particles_in_air">
          <type kind="kind_phys" units="kg-1">real</type>
       </standard_name>
-      <standard_name name="surface_dimensionless_exner_function">
+      <standard_name name="exner_function_wrt_surface_pressure"
+                   long_name="exner function w.r.t. surface pressure, (p/ps)^(Rd/cp)">
          <type kind="kind_phys" units="1">real</type>
       </standard_name>
-      <standard_name name="dimensionless_exner_function_at_surface_adjacent_layer">
+      <standard_name name="exner_function_at_surface_adjacent_layer"
```
nusbaume, line 3772: I believe this should be `wrt`, not `at` (as we are specifying what `p0` is, not just the value of the exner function at a given verical level):

```suggestion
      <standard_name name="exner_function_wrt_surface_adjacent_layer_pressure"
```

standard_names.xml
```diff
@@ -3178,19 +3758,23 @@
...
+                   long_name="exner function (p/p0)^(Rd/cp), where p0 is the pressure at the surface adjacent layer">
          <type kind="kind_phys" units="1">real</type>
       </standard_name>
-      <standard_name name="dimensionless_exner_function_at_interfaces">
+      <standard_name name="exner_function_at_interfaces"
+                   long_name="exner function (p/p0)^(Rd/cp), where p0 is the pressure at vertical layer interfaces">
```
nusbaume, line 3777: I believe this long name should be changed slightly to:

```suggestion
                   long_name="exner function (p/p0)^(Rd/cp), where p is the pressure at vertical layer interfaces">
```

nusbaume (approved):
Great work @mkavulich!  I had a couple of change requests, but nothing that requires a full re-review.

standard_names.xml
climbfuji: Can we talk about this at the next standard names meeting, please? My immediate reaction is to stick with the basic ASCII characters and nothing else, but maybe I can be convinced otherwise ;-) At the very least, we need to run a test with CCPP (SCM, UFS, ...) if the parsers (prebuild and capgen) can handle those characters.

standard_names.xml
climbfuji: ok

standard_names.xml
mkavulich: From browsing the UFS code, this variable is specifically being used for the exner function at the surface adjacent layer, with respect to surface pressure. So I believe the existing name is correct.

Reference, for example: https://github.com/NCAR/ccpp-physics/blob/main/physics/SFC_Models/Land/Noahmp/noahmpdrv.F90#L546

standard_names.xml
mkavulich: Same here, `p0` is still the surface pressure, it's `p` that's at interfaces. P0 is defined as 1000 hPa in [GFS_rrtmg_pre.F90](https://github.com/NCAR/ccpp-physics/blob/main/physics/Interstitials/UFS_SCM_NEPTUNE/GFS_rrtmg_pre.F90#L387)

It could probably be re-worded to be more clear in this case, what do you think about this suggestion:

> Exner function (p/p0)^(Rd/cp) at vertical layer interfaces, where p0 is 1000 hPa

nusbaume: Hi @mkavulich, I think I am slightly confused by your response here?  I guess I was assuming that that if this variable is defined on the interfaces, then the only quantity that will actually vary on the interfaces is just the air pressure (`p`), and thus the update to the long_name would make that clear.  Maybe that's not the case though?

standard_names.xml
nusbaume: Hi again, I realized I misunderstood the code you were pointing me to.  My bad!  I think in this case the `standard_name` is correct (assuming it's using a globally constant `p0` value), but the `long_name` should be updated to the following:

`exner function (p/p0)^(Rd/cp), where p is the pressure at the surface adjacent layer`

Otherwise I think the long name implies that p0 varies in time and space, which it sounds like is not the case.

standard_names.xml
```diff
@@ -405,16 +891,16 @@
...
       <type kind="kind_phys" units="1">real</type>
     </standard_name>
     <standard_name name="reference_pressure_in_atmosphere_layer_normalized_by_surface_reference_pressure"
                    long_name="Reference pressure in atmosphere layer normalized by surface reference pressure">
       <type kind="kind_phys" units="1">real</type>
     </standard_name>
-    <standard_name name="dimensionless_exner_function"
-                   long_name="exner function">
+    <standard_name name="exner_function"
+                   long_name="exner function, (p/p0)^(Rd/cp)">
```
mkavulich, line 903: Clarify that this is vs constant p0 and the other is variable surface pressure

standard_names.xml
```diff
@@ -3178,19 +3758,23 @@
...
          <type kind="kind_phys" units="kg-1">real</type>
       </standard_name>
-      <standard_name name="surface_dimensionless_exner_function">
+      <standard_name name="exner_function_wrt_surface_pressure"
+                   long_name="exner function w.r.t. surface pressure, (p/ps)^(Rd/cp)">
          <type kind="kind_phys" units="1">real</type>
       </standard_name>
-      <standard_name name="dimensionless_exner_function_at_surface_adjacent_layer">
+      <standard_name name="exner_function_at_surface_adjacent_layer"
+                   long_name="exner function (p/p0)^(Rd/cp), where p0 is the pressure at the surface adjacent layer">
```
mkavulich, line 3773: Include correct P0s here

standard_names.xml
mkavulich: From today's discussion, we will restrict standard names (including long names) to ASCII character set.

standard_names.xml
mkavulich: We will use "for coupling" suffix, and include a detailed description of what that means (pending further potential discussion)

standard_names.xml
mkavulich: We will endorse the "case insensitivity" rule.

standard_names.xml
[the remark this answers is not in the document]
mkavulich: Clarify that these are "preferred units" that indicate dimensionality and should be used if possible to avoid conversion overhead. Will include these in all sections (with "1" if units don't make sense)

standard_names.xml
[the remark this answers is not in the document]
mkavulich: We discussed at today's meeting that "weight" is a synonym for "scaling factor" in all existing uses, so we will convert these to "scaling factor".

tools/write_standard_name_table.py
```diff
@@ -156,10 +160,10 @@ def convert_xml_to_markdown(root, library_name, snl):
...
                     snl.write('{}\n'.format(sec_comment))
                     sec_comment = ''
-                # end if
-            # end while
-        # end if
-        for std_name in section:
+        for std_name in sec:
+            if std_name.tag == 'section':
+                parse_section(snl, std_name, level + '#')
+                continue
```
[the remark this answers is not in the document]
gold2718: It doesn't look too indenty to me but I don't think there is a standard here (even Linus breaks his "... if you need more than 3 levels of indentation, you're screwed anyway, and should fix your program." rule).

standard_names.xml
```diff
@@ -1,4 +1,490 @@
...
+      <standard_name name="diffuse_nir_shortwave_flux"
+                     long_name="diffuse near-infrared shortwave flux">
+      </standard_name>
+      <standard_name name="diffuse_shortwave_albedo"
+                     long_name="diffuse_shortwave_albedo">
+      </standard_name>
+      <standard_name name="diffuse_uv_and_vis_shortwave_flux"
+                     long_name="diffuse ultraviolet and visible shortwave flux">
+      </standard_name>
+      <standard_name name="diffuse_visible_albedo"
```
gold2718, line 315: Is this one getting changed to `vis`?

standard_names.xml
```diff
@@ -1,4 +1,500 @@
...
+      <standard_name name="direct_nir_albedo"
+                     long_name="direct_nir_albedo">
+      </standard_name>
+      <standard_name name="direct_nir_shortwave_flux"
+                     long_name="direct_nir_shortwave_flux">
+      </standard_name>
+      <standard_name name="direct_uv_and_vis_shortwave_flux"
+                     long_name="direct_uv_and_vis_shortwave_flux">
+      </standard_name>
+      <standard_name name="direct_visible_albedo"
```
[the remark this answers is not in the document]
gold2718: These don't seem to have been changed.

standard_names.xml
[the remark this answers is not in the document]
gold2718: To me, this feels very imprecise and not very much in the spirit of all the exacting language we have been working on. This is more to the community but could this be spelled out in the name by separating sources/sinks? `dycore` is a pretty different source than `emission`.

gold2718 (changes requested):
This looks better but not all the issues are resolved (e.g., `vis` to `visible`), at least as of cd5ba374b7b8e0f106783eab93f31bae7479978a.

standard_names.xml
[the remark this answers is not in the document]
mkavulich: @gold2718 Since this is a multi-round effort I am trying not to snowball additional disparate changes into this PR; I have noted this as a "to-do" for the next branch. Sorry I wasn't more clear in my initial reply.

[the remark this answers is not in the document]
mkavulich: Since this wasn't all that big of a change, I went ahead and applied it here.

standard_names.xml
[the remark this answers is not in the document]
mkavulich: I think this needs further discussion/investigation since this is an existing long-standing name;  would it be all right if we pushed this to the next round of changes?

standard_names.xml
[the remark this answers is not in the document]
gold2718: Thanks! What would help in these situations if there were one or more issues open with the "to-do" list so I could cross-check before complaining.

standard_names.xml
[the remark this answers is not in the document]
gold2718: Sure, see #90.

gold2718:
Will #86 be merged in before final review?

mkavulich:
@gold2718 I did not plan to resolve conflicts between the main branch and release branch at this time, since the changes are already so divergent. But this is a good idea; I will make a follow-up PR right now to bring in the relevant changes since this branch was cut.

I thought I had responded to your latest comments but I guess I forgot; I made the changes you requested re: vis to visible, was there anything else that needed attention or are you okay with this being merged?

svahl991:
This PR changes many of the names that we just adapted in JEDI last fall during a code sprint that took considerable effort. Before merging this, can we create a tag or version of this repository that JEDI can reference?

standard_names.xml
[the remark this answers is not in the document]
gold2718: Please open an issue describing this change.

gold2718:
> @gold2718 @climbfuji Thanks for your thorough reviews. I've addressed most of your comments/suggestions, and those that I haven't I replied with a follow-up comment or question. Let me know if anything else needs resolving.

Please just open issues for any changes that need changing (based on comments on this PR) so that they are not lost. I do not see one for the `weight` to `scaling_factor` change.

gold2718:
> This PR changes many of the names that we just adapted in JEDI last fall during a code sprint that took considerable effort. Before merging this, can we create a tag or version of this repository that JEDI can reference?

No objections here. `v0.1.2`?

mkavulich:
> This PR changes many of the names that we just adapted in JEDI last fall during a code sprint that took considerable effort. Before merging this, can we create a tag or version of this repository that JEDI can reference?

@svahl991 Just a note, this PR is not being merged to `main`, but to a branch `release/v1` as part of a multi-stage effort to make the names and rules consistent. We will not be applying these changes to the main branch until there has been substantial input from all interested parties.

We had discussed creating a tag back in the fall for JEDI's "stable" version but last time we had that discussion there were still some changes being debated. Is the current state of main the preferred version to make this tag? Or would it be a previous hash?

mkavulich:
@gold2718 I have opened some issues that I think covers all of your concerns; I have been tracking them in a google doc but making sure they are tracked here as well is a good idea.

svahl991:
> Just a note, this PR is not being merged to main, but to a branch release/v1 as part of a multi-stage effort to make the names and rules consistent.

Thanks @mkavulich . That's an important detail I missed. So, to make sure I understand:

1. The `main` branch is still the "official" ESM (formerly CCPP) name for the time being.
2. The `release/v1` branch is building up a major re-work of the rules and names, but is not yet "official".
3. The plan is to build this re-work up in the `release` branch, get consensus, and merge it into `main` at some point in the future.

I apologize if I missed and/or forgot past communications about this.

gold2718 (approved):
All looks good now, thanks!
(and thanks to @mkavulich for getting all those issues open)

Metadata-standard-names.md
```diff
@@ -1727,13 +1926,13 @@ Variables related to the compute environment
     * `real(kind=kind_phys)`: units = K
 * `volumetric_soil_moisture_between_soil_bottom_and_water_table`: Volumetric soil moisture between soil bottom and water table
     * `real(kind=kind_phys)`: units = m3 m-3
-* `water_vapor_mixing_ratio_wrt_moist_air_at_2m`: mixing ratio of the mass of water vapor to the mass of moist air, at two meters above surface
+* `water_vapor_mixing_ratio_wrt_moist_air_at_2m`: Specific humidity (water vapor mass mixing ratio with respect to moist air) at two meters above surface
```
svahl991, line 1929: We have found in our work on JEDI that the term "specific humidity" does not have an internationally agreed-upon meaning. I suggest not using it at all to avoid confusion. Or, if it must be used, to qualify it. (When the UKMO uses the term "specific humidity" they mean `water_vapor_mixing_ratio_wrt_moist_air_and_condensed_water`, which is two lines below this one.)

mkavulich:
@svahl991 Your understanding is correct. This re-factor is part of a lead-up to standing up a more formal governance outside of the CCPP framework's purview; we don't yet have anything official but I have been putting out feelers to ensure that all interested groups are represented. As an active contributor from JCSDA, I'll certainly be sure you're in the loop when we start scheduling meetings.

With regards to specific humidity, you'll note that this PR removes `specific_humidity` from all standard names, exactly because of the problems you mentioned (conflicting/non-standardized definitions), see [here](https://github.com/ESCOMP/ESMStandardNames/pull/87/files#diff-bb599d1972097b4e19b15ce1f799e76cf32576a4281e742ddd07ff578978482bL1719) for example. But because it is a common term, we decided it should stay in long_name entries where relevant, with a parenthetical qualifier of the exact definition as you see here. I could see where it might still be an issue though, so I will open an issue where we can discuss this in more detail.

PR #106, opened 2025-03-26
Cherry-picked from V1 branch: fix misspellings in Standard Names

mkavulich:
There are many misspellings in the existing `main` branch, not only in the `long_name` descriptions but in the `standard_name` entries themselves. This PR addresses the ones I have identified so far.


gold2718 (approved):
Great catches! I'm a bit curious as to how you found these. Is there some way we can semi-automate this process?

mkavulich:
@gold2718 This was originally done a few months ago, but I believe I did it with vim's built-in spell-checker. I'm not sure of a good way to automate the process given the number of field-specific terms that would trip any standard spell check, but I'm open to suggestions if anyone has them.

nusbaume (approved):
Thanks @mkavulich!

Issue #103, opened 2025-03-21
Rename `long_name` attribute to `description`


mkavulich:
The term `long_name` comes from the repository's CCPP provenance and is somewhat confusing and ambiguous. Since this is essentially a field for including a detailed description of a variable's use, it is much more clear and straightforward to rename this attribute `description`.

gold2718:
The term, `long_name`, comes from the [CF conventions](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.12/cf-conventions.pdf), see section 1.4. I think we should keep it.

climbfuji:
I agree with @gold2718, I prefer keeping `long_name` over other descriptors.

cacraigucar:
One "nuance" which was discussed at our CCPP meeting was that while `standard_name` is required to match exactly what is in the dictionary,  `long_name` is not required to match across models.  By its inclusion in the dictionary, users may feel compelled to include the exact `long_name` in their models as well since the framework uses the same name in .meta files.  We discussed how we do NOT want to change `long_name` in the Framework.   If the dictionary does not give "guidance" on how to populate the `long_name` field, then users will be more comfortable using names which better describe the variables within their particular models. 

The dictionary is describing what the variable "is" to aid users in discovering their correct variable.  So the current `long_name` field is more accurately a `description`. 

When fields are being used within the framework .meta files, the `long_name` there IS adhering to the CF conventions.

I personally believe that the `description` in the dictionary and `long_name` in .meta files are two separate quantities and the place where we need to adhere to the CF conventions is in the Framework itself.

gold2718:
I find @cacraigucar's argument fairly persuasive. I am okay with this but have a couple of requests.

1. The distinction from and relationship to the `long_name` metadata argument needs to be clearly documented in the preamble.
2. Every entry gets a description that is -- descriptive.

climbfuji:
> One "nuance" which was discussed at our CCPP meeting was that while `standard_name` is required to match exactly what is in the dictionary, `long_name` is not required to match across models. By its inclusion in the dictionary, users may feel compelled to include the exact `long_name` in their models as well since the framework uses the same name in .meta files. We discussed how we do NOT want to change `long_name` in the Framework. If the dictionary does not give "guidance" on how to populate the `long_name` field, then users will be more comfortable using names which better describe the variables within their particular models.
> 
[...]

Issue #95, opened 2025-03-04
[release/v1] Define and include more standard abbreviations in the list of Acronyms, Abbreviations, and Aliases


mkavulich:
Add a rule that abbreviations and acronyms should be included in the standard name where appropriate, but expanded to the full name in the long_name description.

Specific examples:
 - MYNN --> Mellor-Yamada-Nakanishi-Niino (scheme)
 - MYJ --> Mellor-Yamada-Janjic (scheme)
 - PBL --> planetary boundary layer
 - GWD --> gravity wave drag
 - vis --> visible

Add more examples to this issue as needed

mkavulich:
Even more:

CS --> Chikira-Sugiyama
EDMF --> eddy-diffusivity mass flux
pdf --> probability density function
RRTMG --> Rapid Radiative Transfer Model for global climate model (GCM) applications
RRTMGP --> RRTMG re-written for parallelization
sgs --> subgrid-scale
SAS --> Simplified Arakawa-Schubert
sw --> shortwave (used in some names but not all)

mkavulich:
Even more:

cnvc90 --> GFS Convective Cloud Diagnostics
gfdl --> Geophysical Fluid Dynamics Laboratory
gsl --> Global Systems Lab
lsm --> land surface model
mcica --> Monte-Carlo Independent Column Approximation
noahmp --> Noah land surface model with multiparameterization options
nrl --> Naval Research Lab
nsstm --> GFS near-surface sea temperature scheme
rte --> Radiative Transfer for Energetics
ugwp --> Unified Gravity Wave Physics

mkavulich:
Resolved by #116

PR #116, opened 2025-06-05
[release/v1] Fourth major update in v1 branch: Rename `long_name` --> `description`, expand list of abbreviations, update descriptions

### LINKED ISSUE #93: [release/v1] Expand on long_name descriptions for base names
As discussed [here](https://github.com/ESCOMP/ESMStandardNames/pull/87#discussion_r1898132767)

mkavulich:
## Description
This PR attempts to address the following issues:
  - #95: Standardize and define more abbreviations used in standard_names
  - #103: renaming `long_name` --> `description` to avoid confusion with the `long_name` attribute in CCPP/CF conventions.

In addition, this PR clarifies that the description field should be unique, and adds a check for this (using a re-written tools/check_xml_unique.py script with expanded functionality)

## Issues

Resolves #93, #95, #103


StandardNamesRules.rst
```diff
@@ -165,23 +165,27 @@ Technical specifications
 ========================
 
 #. The standard name dictionary consists of a number of individual XML elements:
-   one "standard_name" element for each entry. A standard name entry consist of a "name" attribute
-   that represents the variable name, and (optionally) a "long_name" attribute that gives
-   a detailed description of what that name represents. The standard_name XML entry also contains a nested
-   "type" entry, indicating the data type that a standard_name should represent, and as attributes the
+   one ``standard_name`` element for each entry. A standard name entry consist of a ``name`` attribute
```
climbfuji, line 168: Can you add a sentence to this section to explain what the long name is as opposed to the description, where it is being used, please? It's entirely unclear from this paragraph that the description is for the purpose of the ESMStandardNames repo only, whereas the long_name is for the actual metadata tables only.

Metadata-standard-names.md
climbfuji, line 1: I didn't really review this one, since you created it from the XML I assume.

StandardNamesRules.rst
```diff
@@ -646,10 +656,14 @@ Acronyms, Abbreviations, and Aliases
 +---------------------+---------------------------------------------------------+
 | nir                 | near-infrared part of the EM spectrum (radiation)       |
 +---------------------+---------------------------------------------------------+
```
climbfuji, line 658:
```suggestion
+---------------------+---------------------------------------------------------+
| nrl                 | Naval Research Lab                                      |
+---------------------+---------------------------------------------------------+
```
I checked the xml below and nrl is being used.

climbfuji (changes requested):

StandardNamesRules.rst
mkavulich: @climbfuji I'm having some trouble coming up with some satisfying wording here. Right now I think the mention of "long_name" adds more confusion than it resolves. Would it make sense to just remove the reference to "long_name" all together, since with this proposed change it would no longer be a part of the Standard Names? That way we can just say the following, which I believe is clear and unambiguous:

>A standard name entry consist of a ``name`` attribute that represents the variable name, and (optionally) a ``description`` attribute that gives a detailed description of what that name represents. Note that the ``description`` field is only provided for information and disambiguation only (though it should be unique), and does not need to be included for individual implementations of the standard names.

If you have other suggestions (general or specific) I welcome them, because after thinking for a while I'm a bit lost on the best way to word this bit.

nusbaume (approved):
Everything looks great to me now.  Thanks!

mkavulich:
@climbfuji In an attempt to address your comments about pointing to CCPP docs description for `long_name`, I was sent down a bit of a rabbit hole...it turns out, we don't ever actually define `long_name` in the CCPP docs, nor do we define many other metadata entries. So I opened a new PR (https://github.com/NCAR/ccpp-doc/pull/77) to add these descriptions, and in doing so, updated other parts of that section that were badly out of date. Take a look over there, and if that's acceptable as a description I'll add a link here.

PR #134, opened 2026-01-07
Fix convert_xml_to_yaml() function to properly handle nested sections

### LINKED ISSUE #132: Missing names in Metadata-standard-names.yaml?
Currently in Metadata-standard-names.md and standard_names.xml, I see entries for `number_of_openmp_threads` and for `ccpp_thread_number`. However, in Metadata-standard-names.yaml, I see only `number_of_openmp_threads`. Is this intentional or is it a bug?

mkavulich:
## Description
The current version of the `tools/write_standard_name_table.py` script does not appropriately read nested section elements from the XML when writing a YAML output. I had ChatGPT fix the function to appropriately account for this, then reviewed the results and made some manual tweaks to make the output neater.

Links to ChatGPT output: 
 - https://chatgpt.com/s/t_695daa1689048191bd3c2c79684750cd
 - https://chatgpt.com/s/t_695dacf909dc8191969fa261c0287fbc

## Issues
Fixes #132

### DIFF
--- a/tools/write_standard_name_table.py
+++ b/tools/write_standard_name_table.py
@@ -245,20 +245,19 @@ def parse_section_for_yaml(section):
             stdn_description = standard_name_to_description(sdict)
 
         std_type = std_name.find('type')
-        if std_type is None:
-            continue
 
         std_name_data = OrderedDict()
         std_name_data['name'] = stdn_name
         std_name_data['description'] = stdn_description
-        std_name_data['type'] = std_type.text
-        std_name_data['kind'] = std_type.get('kind')
+        if std_type is not None:
+            std_name_data['type'] = std_type.text
+            std_name_data['kind'] = std_type.get('kind')
 
-        units = std_type.get('units')
-        try:
-            std_name_data['units'] = int(units) if units is not None else None
-        except (ValueError, TypeError):
-            std_name_data['units'] = units
+            units = std_type.get('units')
+            try:
+                std_name_data['units'] = int(units) if units is not None else None
+            except (ValueError, TypeError):
+                std_name_data['units'] = units
 
         sec_data['standard_names'].append(std_name_data)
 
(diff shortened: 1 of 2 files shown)


tools/write_standard_name_table.py
```diff
@@ -210,48 +210,72 @@ def convert_xml_to_yaml(root, library_name, yaml_file):
...
+        try:
+            std_name_data['units'] = int(units) if units is not None else None
+        except (ValueError, TypeError):
+            std_name_data['units'] = units
+
+        sec_data['standard_names'].append(std_name_data)
+
+    # If no standard names in section, delete key/value pair
+    if not sec_data['standard_names']:
+        del sec_data['standard_names']
```
nusbaume: Should we print another warning here?

mkavulich: I don't think a warning is necessary here...this represents the case of a section that contains only subsections, and not any standard names. If not included, an empty list of standard names appeared in the resulting YAML for these sections, which isn't desirable.

mkavulich:
@climbfuji Pinging to make sure you agree with [this change](https://github.com/ESCOMP/ESMStandardNames/pull/134/changes/3032cf48d1527bd0874dc4aaa8a5f3fa2deac624) for standard names with no type.

climbfuji:
> @climbfuji Pinging to make sure you agree with [this change](https://github.com/ESCOMP/ESMStandardNames/pull/134/changes/3032cf48d1527bd0874dc4aaa8a5f3fa2deac624) for standard names with no type.

Yes, thanks!

tools/write_standard_name_table.py
nusbaume: Ahh, that make sense, thanks for the explanation!  In that case I am fine to leave this as-is then (and will go ahead and mark this conversation as resolved).

PR #124, opened 2025-08-27
[release/v1] Fifth major update in v1 branch: Better information on units, remove `kind`, clarify disallowed terms

mkavulich:
## Description
This PR addresses the following issues:
 -  #94 Include more information on the role of "units"
 - #102 Remove "kind" from standard name entries 
 - Partially addresses #105: this PR removes the ambiguous term `amount` replacing it with more specific terms where needed, and includes a new section on "disallowed terms" that includes `specific_humidity` and `amount`.

In addition, this PR organizes the "constants" section alphabetically, and replaces the too-specific `dry_air_density` base name with `dry_air`, and change `dry_air_density` to `density_of_dry_air` to better fit with naming conventions in the updated rules.

## Issues
 - Resolves #94
 - Resolves #102

PR #142, opened 2026-03-25
Remove version information from schema filename, remove obsolete features

mkavulich:
## Description
Because versioning is handled with GitHub tags, and we do not support multiple schema versions for a given tag, it is confusing and unnecessary to include the version number in the filename of the schema file. Therefore this PR renames  
standard_names_v1_0.xsd --> standard_names.xsd, and removes the python tooling needed to resolve the schema filename based on version.

In the process of making these updates, I noticed there is a bunch of logic for very old, unsupported python versions, and some remaining logic related to `kind` information that was removed in #124. I removed this unnecessary logic.

## Issues
Resolves #141


gold2718 (approved):
Approving although I'm sorry to see the Python2 go, it helps to confuse the bots :)

climbfuji (commented):
Do you know why only `Metadata-standard-names.yaml` got updated, but not `Metadata-standard-names.md`?

climbfuji (approved):

PR #143, opened 2026-04-09
Add field for CF equivalent name

### LINKED ISSUE #108: [release/v1] Add a new field for equivalent CF standard name?
Since we have decided to allow divergence from the original CF standard names, but there is still a large amount of overlap in both the names and the applications, should we attempt to introduce some inter-compatibility for the equivalent CF standard names? It seems like this could be useful down the road for applications that have or deal with a mix of the two naming conventions.

As an example of what I'm thinking, an optional xml entry "cf_name" could exist alongside "standard_name":

```
    <standard_name name="air_pressure_at_surface"
                   cf_name="surface_air_pressure" long_name="Air pressure at local surface">
      <type kind="kind_phys" units="Pa">real</type>
```

While there's no way to ensure that the association is correct aside from manual review, this field could be automatically checked against the current CF standard names table (https://cfconventions.org/Data/cf-standard-names/current/src/cf-standard-name-table.xml) to ensure the cf_name is valid.

mkavulich:
## Description
This PR updates the XML, schema file, and Metadata files to include and "Equivalent CF name" field, to make it easier for code bases such as JEDI that need to interface with both CF standard names and ESM standard names.

In this process I also found and removed some duplicate names, and moved some names to better sections

## Issues
Resolves #108

### DIFF
--- a/standard_names.xml
+++ b/standard_names.xml
@@ -366,6 +366,7 @@
         <type units="K">real</type>
       </standard_name>
       <standard_name name="virtual_temperature" description="The theoretical temperature of dry air that would have the same density as moist air">
+        <cfname>virtual_temperature</cfname>
         <type units="K">real</type>
       </standard_name>
       <standard_name name="water_vapor" description="Water in the gaseous phase">
@@ -373,6 +374,7 @@
       <standard_name name="wind" description="Movement of air with a net displacement">
       </standard_name>
       <standard_name name="wind_speed" description="Speed of moving air">
+        <cfname>wind_speed</cfname>
         <type units="m s-1">real</type>
       </standard_name>
       <standard_name name="wind_stress" description="Shear stress exerted by wind parallel to the surface">
@@ -887,6 +889,7 @@
       <type units="m s-1">real</type>
     </standard_name>
     <standard_name name="x_wind" description="Horizontal wind in a direction perpendicular to y_wind">
+      <cfname>x_wind</cfname>
       <type units="m s-1">real</type>
     </standard_name>
     <standard_name name="x_wind_at_surface_adjacent_layer">
@@ -899,6 +902,7 @@
       <type units="m s-1">real</type>
     </standard_name>
     <standard_name name="y_wind" description="Horizontal wind in a direction perpendicular to x_wind">
+      <cfname>y_wind</cfname>
       <type units="m s-1">real</type>
     </standard_name>
     <standard_name name="y_wind_at_surface_adjacent_layer">
@@ -3595,6 +3599,7 @@
       <type units="frac">real</type>
     </standard_name>
     <standard_name name="vegetation_area_fraction">
+      <cfname>vegetation_area_fraction</cfname>
       <type units="fraction">real</type>
     </standard_name>
     <standard_name name="vis_albedo_strong_cosz" description="albedo for visible radiation with strong dependence on cosine of the zenith angle">
@@ -3604,6 +3609,7 @@
       <type units="fraction">real</type>
     </standard_name>
     <standard_name name="volume_fraction_of_condensed_water_in_soil">
+      <cfname>volume_fraction_of_condensed_water_in_soil</cfname>
       <type units="fraction">real</type>
     </standard_name>
     <standard_name name="volume_fraction_of_frozen_soil_moisture_for_lsm" description="Volume fraction of frozen soil moisture for land surface model">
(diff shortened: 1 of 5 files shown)


climbfuji (approved):
At first, I was confused of seeing `cfname` entires when they were identical to the ESM standard name, but I realized that this makes sense. Thanks for weeding through the CF names and adding the matches here.

svahl991:
So let me confirm my understanding with this PR:

- Not every ESM name has an equivalent CF name
- Not every CF name has an equivalent ESM name
- Where there is an equivalence, it must be one-to-one

Is there code in place somewhere to enforce the last statement? Put another way, is there anything (other than human diligence) to prevent someone from accidentally pointing a new ESM name at a CF name that is already pointed to by another ESM name?

mkavulich:
@svahl991 You bring up a good point, we should have an automated check for duplicate "cfname" elements, since there should never be duplicates in our dictionary unless we have messed something up. I was going to add a quick check in this PR, but it turns out our current tests aren't in very good shape (#144), so I would like to open a new PR to address this.

mkavulich:
@svahl991 I have opened #145 that includes a check for duplicate CF names, which should run automatically with each PR.

And with regards to your other questions, I think your understanding is in line with my thinking. The CF names do not cover all our use cases, and we have not "automatically" pulled in all CF names (though we likely would if they were needed).

mkavulich:
I'll also note for the record, most ESM names are identical to the equivalent CF names, but a few are different because of divergence decisions we've made in the past few years (e.g. `air_potential_temperature` vs. `potential_temperature_of_air`) or other reasons that are unclear to me (e.g. `stratiform_cloud_area_fraction_in_atmosphere_layer` vs. `nonconvective_cloud_area_fraction_in_atmosphere_layer`). It is possible that some of my equivalence additions here are incorrect due to subtle definition differences but I've done my best to ensure they are used the same, at least in CCPP.

standard_names.xml
```diff
@@ -3328,26 +3384,33 @@
...
       <type units="1">real</type>
     </standard_name>
     <standard_name name="leaf_mass_content">
       <type units="g m-2">real</type>
     </standard_name>
     <standard_name name="lwe_snowfall_rate" description="Liquid water equivalent snowfall rate">
+      <cfname>lwe_snowfall_rate</cfname>
       <type units="mm s-1">real</type>
     </standard_name>
     <standard_name name="lwe_surface_snow" description="Liquid water equivalent surface snow">
```
nusbaume, line 3408: I realize this isn't directly related to this PR, but is this variable any different from `lwe_thickness_of_surface_snow`, which is listed farther down?

tools/write_standard_name_table.py
```diff
@@ -184,9 +184,11 @@ def parse_section(snl, sec, level='##'):
             stdn_description = standard_name_to_description(sdict)
         # end if
         snl.write("* `{}`: {}\n".format(stdn_name, stdn_description))
-        # Should only be a type in the standard_name text
+        # Should only be type or cfname as subelements of standard_name
         for item in std_name:
-            if item.tag == 'type':
+            if item.tag == 'cfname':
+                snl.write(f"    * Equivalent CF name: {item.text}\n")
```
nusbaume: I could definitely be wrong as I am not a Markdown expert, but should this have quotations marks like the standard name and type outputs?  For example:




```suggestion
                snl.write(f"    * `Equivalent CF name`: {item.text}\n")
```

nusbaume (approved):
Thanks for bringing this in @mkavulich!  I had a couple questions which came up while reviewing the PR, but otherwise everything looks good to me.

svahl991 (approved):

standard_names.xml
mkavulich: Good catch! This appears to be an original duplicate from CCPP. There is some room here for two variables, since  LWE could be measured as mass per unit area, or thickness, but both of these variables have units of length, implying the latter measure. I'll add this to a list of potential duplicates to be addressed later.

tools/write_standard_name_table.py
mkavulich: I actually think it's better to put the ticks around the name rather than the label here; that way the ESM standard name and CF standard name are formatted the same way

mkavulich: I'll go ahead and merge this now to move things along, but if you disagree or have an alternative suggestion, let me know and I can change it in the next PR (#145)

nusbaume: @mkavulich your change makes sense to me!  I'm happy for this PR to be merged in as-is now.

PR #98, opened 2025-03-04
Changes after rename of CCPPStandardNames --> ESMStandardNames

mkavulich:
- Remove references to CCPP except where relevant
 - Cherry-pick changes to tools/write_standard_name_table.py from release branch that allows for subsections in standard names and markdown table
 - Group CCPP-specific variables into a dedicated section
 - Change some long_name descriptions to remove CCPP-specific wording
 - A few more details in README.md

PR #145, opened 2026-04-13
Fix problems with tools, simplify and add checks with pylint

### LINKED ISSUE #118: Update python scripts to use f-strings
Currently the python scripts located in the `tools` directory use different methods for including variables within strings, usually using either f-strings or the `format()` string method.

For the sake of readability and coding style uniformity, it might be good to modify all of the scripts to only use f-strings, at least for all situations where that is a valid option.

### LINKED ISSUE #144: Need to fix tests to properly recurse subsections
There are multiple tests/tools that do not properly handle XML subsections, and so are currently only reading/checking the standard names in top-level sections:

 - tools/ccpp_meta_stdname_check.py
 - tools/check_name_rules.py
 - tools/check_xml_unique.py

These should be fixed. Along the way, it's probably good to reduce some of the logic/function duplication between tools which accomplish similar things.

mkavulich:
## Description
In the process of trying to add a new tool to check for redundant CF name elements (see #143), I found that our existing tools suffer from a number of issues:

 - Several tools were not updated appropriately to deal with nested sections (#98) so they are not properly checking all the names in the dictionary
   - This meant that there were several redundant standard names that had to be removed
 - A bad merge at some point re-introduced some non-working code from the framework in tools/write_standard_name_table.py; since doctests are not used in this repository, I removed those tests and the un-exercised logic
 - Changes in #142 that standardized the spelling of the schema file left a lot of stub logic that should be removed

This PR resolves those problems. Also, I took the opportunity to clean up the existing tools:
 - Make all python scripts pass pylint with a 10/10 score
   - Removed some old format strings, replacing with f-strings (#118)
 - Standardized the format of comments and multi-line strings
 - Standardized how command-line arguments are named and referenced
 - Removed many unnecessary comments

And finally, I added the check for duplicate cfname elements that was the start of this whole process.

## Issues
Resolves #144, partially resolves #118


tools/lib/xml_tools.py
```diff
@@ -39,129 +39,96 @@ def call_command(commands, logger, silent=False):
...
-            outstr += "{}".format(err.output)
-            raise RuntimeError(outstr)
-        # end if
-    # end of try
+            outstr += f"{err.output}"
+            raise RuntimeError(outstr) from err
     return result
 
-###############################################################################
-def find_schema_file(schema_root, schema_path=None):
```
mkavulich, line 64: Note: this tool is no longer needed since we no longer have the schema version in the filename (#142)

tools/lib/xml_tools.py
```diff
@@ -39,129 +39,96 @@ def call_command(commands, logger, silent=False):
...
         logger.warning(lmsg.format(filename))
-    # end if
     return True # We could not check but still need to proceed
 
 ###############################################################################
 def read_xml_file(filename, logger=None):
 ###############################################################################
     """Read the XML file, <filename>, and return its tree and root"""
-    if os.path.isfile(filename) and os.access(filename, os.R_OK):
-        file_open = (lambda x: open(x, 'r'))
```
mkavulich, line 133: This logic was simplified as a suggestion from pylint

tools/ccpp_meta_stdname_check.py
```diff
@@ -75,46 +51,19 @@ def parse_arguments():
...
     #Parse Argument inputs
     args = parser.parse_args()
 
     return args.metafile_loc, args.stdname_dict
 
-#+++++++++++++++++++++++++++++++++++++++++++++++++++++++++
-#Function to extract standard names from element tree root
-#+++++++++++++++++++++++++++++++++++++++++++++++++++++++++
-
-def get_dict_stdnames(xml_tree_root):
```
mkavulich, line 95: Renamed this function to `get_standard_names_as_set` to avoid confusion (it returns a set, not a dict)

tools/ccpp_meta_stdname_check.py
```diff
@@ -276,90 +202,56 @@ def print_missing_names(missing_names_dict):
...
             print(f"    - {stdname}")
-        #End for
-
-    #End for
 
     print("\n#######################")
 
 #+++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
-############
```
mkavulich, line 287: This logic isn't really changed all that much, just moved into function `main_func()`; choosing the "Hide whitespace" option will make this bit easier to review.

tools/lib/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""Shared utilities for ESM Standard Names tools."""
```
mkavulich, line 1: This allows us to import these functions directly from tools in the directory above; note the simplified logic in imports at the top of those scripts.

tools/check_name_rules.py
```diff
@@ -31,33 +20,25 @@ def main():
...
     stdname_file = os.path.abspath(args.standard_name_file)
-    tree, root = read_xml_file(stdname_file)
+    _, root = read_xml_file(stdname_file)
 
     # Validate the XML file
-    schema_name = os.path.basename(stdname_file)[0:-4]
     schema_root = os.path.dirname(stdname_file)
-    schema_path = os.path.join(schema_root,schema_name)
-    schema_file = find_schema_file(schema_path)
-    if schema_file:
```
mkavulich, line 47: Removed several of these schema file checks, since `validate_xml_file` function already checks for its existence

tools/check_name_rules.py
```diff
@@ -70,9 +51,6 @@ def main():
         if violations:
             violators[sname] = violations
 
-    if violators:
```
mkavulich, line 73: We want to check all violations, not just the first set to fail

tools/check_xml_unique.py
```diff
@@ -42,38 +27,37 @@ def parse_command_line(args, description):
...
-            raise ValueError(f"Invalid standard names file, {stdname_file}")
-    else:
-        raise ValueError(f'Cannot find schema file, {schema_name}')
+    schema_path = os.path.join(schema_root,"standard_names.xsd")
+    validate_xml_file(stdname_file, schema_path, logger=None, error_on_noxmllint=True)
 
     #get list of all standard names
     all_std_names = []
-    for name in root.findall('./section/standard_name'):
+    for name in root.findall('.//standard_name'):
```
mkavulich, line 46: The change that started this whole rabbit hole: we need to check for ALL standard_name entries, not just the top-level ones

tools/check_xml_unique.py
```diff
@@ -88,30 +72,16 @@ def main_func():
     if len(dup_std_names)>0:
         print(f'The following duplicate {args.field} entries were found:')
         for dup in dup_std_names:
-            rm_elements = root.findall(f'./section/standard_name[@{args.field}="{dup}"]')[1:]
+            if args.field == 'cfname':
+                rm_elements = root.findall(f'.//standard_name[cfname="{dup}"]')[1:]
+            else:
+                rm_elements = root.findall(f'.//standard_name[@{args.field}="{dup}"]')[1:]
             print(f"{dup}, ({len(rm_elements)} duplicate(s))")
-        if args.overwrite:
```
mkavulich, line 93: I made a judgement call here that we always want to manually resolve duplicate entries. Doing it automatically seems prone to error (e.g. they were intended to be separate entries but had the same standard name due to a typo).

tools/write_standard_name_table.py
```diff
@@ -70,25 +57,9 @@ def convert_text_to_link(text_str):
     return link_str
 
 ########################################################################
-def standard_name_to_description(prop_dict, context=None):
+def standard_name_to_description(prop_dict):
 ########################################################################
-    """Translate a standard_name to its default description
```
mkavulich, line 75: These checks were remnants from code originally copied from the framework, and have no function here

tools/write_standard_name_table.py
```diff
@@ -98,33 +69,22 @@ def standard_name_to_description(prop_dict, context=None):
...
             description = ''
-        # end if
         # Next, substitute a decimal point for the p in [:digit]p[:digit]
         match = _REAL_SUBST_RE.match(description)
         while match is not None:
             description = match.group(1) + '.' + match.group(2)
             match = _REAL_SUBST_RE.match(description)
-        # end while
     else:
-        description = ''
```
mkavulich, line 109: More logic leftover from ccpp-framework

tools/write_standard_name_table.py
```diff
@@ -288,44 +237,19 @@ def main_func():
...
-    if not schema_file:
-        raise ValueError(f'Cannot find schema file {schema_name}')
-    # end if
-
-    try:
-        emsg = "Invalid standard names file, {}".format(stdname_file)
-        file_ok = validate_xml_file(stdname_file, schema_name,
-                                     None, schema_path=schema_root,
-                                     error_on_noxmllint=True)
-    except ValueError as valerr:
```
mkavulich, line 303: This is convoluted logic to catch an error raised by the `validate_xml_file` function....it's way simpler and better practice to just let the exception be raised and reference the resulting error trace.

mkavulich (commented):
Including some comments to make reviews easier

tools/ccpp_meta_stdname_check.py
```diff
@@ -25,43 +25,19 @@
...
-
-_CURR_DIR = os.path.dirname(os.path.abspath(__file__))
-sys.path.append(os.path.join(_CURR_DIR, "lib"))
-
-#######################################
-#Import needed framework python modules
-#######################################
-
-from xml_tools import read_xml_file
+#Import custom helper functions from lib/ directory
```
climbfuji:
```suggestion
# Import custom helper functions from lib/ directory
```

tools/ccpp_meta_stdname_check.py
```diff
@@ -140,45 +89,27 @@ def find_metafile_stdnames(metafile_obj):
...
             #Attempt to find string index for "equals" sign:
             equals_index = line.find("=")
-
-            #Check if an equals sign actually
-            #exists:
             if equals_index != -1:
 
-                #If so, then extract all text to the right
-                #of the equals sign:
+                #uxtract all text to the right of the equals sign
```
climbfuji:
```suggestion
                # uxtract all text to the right of the equals sign
```
Please add a space for every comment between `#` and the text.

tools/check_xml_unique.py
climbfuji: It triggered a lot of good changes, though!

tools/check_xml_unique.py
```diff
@@ -42,38 +27,37 @@ def parse_command_line(args, description):
...
-            all_std_names.append(name.attrib[args.field])
+            if args.field == 'cfname':
+                # Extract from cfname subelement
+                cfname_elem = name.find('cfname')
+                if cfname_elem is not None and cfname_elem.text is not None:
+                    all_std_names.append(cfname_elem.text.strip())
+                elif args.debug:
+                    print(f"WARNING: no cfname subelement for standard name {name.attrib['name']}")
+            else:
+                # Extract from attribute
```
climbfuji, line 56: Here there's a space. There should one for every comment.

tools/check_xml_unique.py
climbfuji: Agreed

tools/write_standard_name_table.py
climbfuji: Agreed. I see this all the time in the ccpp-framework code.

climbfuji (changes requested):
Great stuff. Thanks for all the cleanup work and improvements.

Only one change request for the format of comments.

mkavulich:
@climbfuji Thanks for the suggestion, I have updated all comments in that file (and fixed a misspelling I noticed)

climbfuji (approved):

svahl991 (approved):

.github/workflows/pull_request_ci.yml
```diff
@@ -8,6 +8,27 @@ on:
       - release/*
 
 jobs:
+  check-tools:
+    name: Pylint all python files under tools/
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
nusbaume, line 16: It probably doesn't matter, but technically `checkout` is now up to `v6`:


```suggestion
        uses: actions/checkout@v6
```

.github/workflows/pull_request_ci.yml
```diff
@@ -8,6 +8,27 @@ on:
...
+
+      - name: Setup Python
+        uses: actions/setup-python@v4
```
nusbaume, line 19: This could also be bumped up to `v5`:


```suggestion
        uses: actions/setup-python@v5
```

tools/lib/xml_tools.py
```diff
@@ -39,129 +39,96 @@ def call_command(commands, logger, silent=False):
...
-        # end if
         result = cproc.returncode == 0
     except (OSError, RuntimeError, subprocess.CalledProcessError) as err:
         if silent:
             result = False
         else:
             cmd = ' '.join(commands)
-            emsg = "Execution of '{}' failed with code:\n"
+            emsg = "Execution of '{cmd}' failed with code: {err.returncode}\n"
             outstr = emsg.format(cmd, err.returncode)
```
nusbaume: I think this is missing the `f` at the start of the string, plus we can just name it "outstr" from the start without the `format` statement:


```suggestion
            outstr = f"Execution of '{cmd}' failed with code: {err.returncode}\n"
```

tools/lib/xml_tools.py
```diff
@@ -39,129 +39,96 @@ def call_command(commands, logger, silent=False):
...
 ###############################################################################
     """
     Find the appropriate schema and validate the XML file, <filename>,
     against it using xmllint
     """
     # Check the filename
     if not os.path.isfile(filename):
-        raise ValueError("validate_xml_file: Filename, '{}', does not exist".format(filename))
-    # end if
+        raise ValueError("validate_xml_file: Filename, '{filename}', does not exist")
```
nusbaume: I think you need an `f` here for this string:


```suggestion
        raise ValueError(f"validate_xml_file: Filename, '{filename}', does not exist")
```

tools/lib/xml_tools.py
```diff
@@ -39,129 +39,96 @@ def call_command(commands, logger, silent=False):
...
-    # end if
-    if not schema_path:
-        # Find the schema file
-        thispath = os.path.abspath(__file__)
-        pdir = os.path.dirname(os.path.dirname(os.path.dirname(thispath)))
-        schema_path = os.path.join(pdir, 'schema')
-    # end if
-    schema_file = find_schema_file(schema_root, schema_path)
-    if not (schema_file and os.path.isfile(schema_file)):
+        raise ValueError("validate_xml_file: Cannot open '{filename}'")
```
nusbaume: I think you need an `f` here as well:


```suggestion
        raise ValueError(f"validate_xml_file: Cannot open '{filename}'")
```

tools/lib/xml_tools.py
```diff
@@ -39,129 +39,96 @@ def call_command(commands, logger, silent=False):
...
-    # end for
     _LOGGER.addHandler(logging.NullHandler())
     try:
         # First, run doctest
         import doctest
         doctest.testmod()
     except ValueError as cerr:
-        print("{}".format(cerr))
-# no else:
+        print("{cerr}")
```
nusbaume: Need an `f` here as well:


```suggestion
        print(f"{cerr}")
```

nusbaume (approved):
Thanks for the tool improvements @mkavulich!  I found some f-strings that are missing the "f", and some possible Github Action versions updates, but otherwise everything looked good to me!

mkavulich:
@nusbaume Thanks for catching that missing f-string notation. I'll make the github action version upgrades in a subsequent PR.

PR #148, opened 2026-06-08
Resolve issues: #24, #64, #82, #118, other fixes

### LINKED ISSUE #24: Inconsistencies between vertical dimension names
We have
```
     <standard_name name="vertical_dimension_for_radiation">
         <type kind="" units="count">integer</type>
      </standard_name>
      <standard_name name="vertical_interface_dimension_for_radiation">
         <type kind="" units="count">integer</type>
      </standard_name>
```
but what used to be `vertical_dimension` (without decorator) is now `vertical_layer_dimension`. This is inconsistent.

It's consistent for `vertical_interface_dimension` and `vertical_interface_dimension_for_radiation`.

### LINKED ISSUE #64: Need to replace the use of "volumetric" with "volume_fraction" when referring to soil quantities
According to the standard name rules, the use of the term `volume_fraction` is preferred over `volumetric`.  This  means that standard names that are currently using `volumetric_soil_X` should be replaced with standard names that use `volume_fraction_of_soil_X`.

Also, assuming the fix for this comes in after PR #60 then it might also help to move all of these land surface standard names to be underneath the new `land_surface` section as well, as opposed to the GFS type(s) section.

### LINKED ISSUE #82: `air_pressure_at_sea_level` or `air_pressure_at_mean_sea_level` or both?
As part of a recent code sprint, we adopted the `CCPP` name `air_pressure_at_sea_level`. This replaced the variable name `surface_pressure_at_mean_sea_level`. It was pointed out after the sprint that the new name omits `mean` and others have questioned why `mean` was removed. Based on post sprint discussion we have had in the UKMO, we feel the omission of `mean` is potentially ambiguous.

In further discussion with others in JEDI, it was also pointed out that according to the CF convention, `_sea_level` is short for `_mean_sea_level`, so they're the same and that `_sea_surface` is the instantaneous sea surface. Its not clear to me why `CF` have two names for the same thing other than one had been established and changing it was difficult (if someone knows the actual reason then please let me know).

While digging into the CCPP I see that this variable was added in 2021 via: https://github.com/ESCOMP/CCPPStandardNames/pull/10 so it is an established CCPP name. The suffex `at_sea_level` is a defined: https://github.com/ESCOMP/CCPPStandardNames/blob/583a37926449110445411b8efb4496b7e1497b93/StandardNamesRules.rst?plain=1#L173 although it does appear that it is only used by `air_pressure`.

Looking at the CCPP PR where this variable was added, there does not appear to be any discussion about the use of `mean` as we have highlighted here.

What do others think about this?

Tagging individuals involved in the original PR for further details and feedback: @cacraigucar @nusbaume @gold2718 @peverwhee 

and those involved in the code sprint for referabce: @svahl991 @mikecooke77 @fmahebert @ncrossette @rtodling @sking112 @danholdaway @travissluka @BenjaminTJohnson @stemiglio

Thanks, Steve

mkavulich:
## Description
Resolving several open issues and items noted on previous PRs:

 - Fix #24
   - Renaming "vertical_layer_dimension" to just "vertical_dimension" for consistency with other dimension names
   - Add a definition of "interface" for clarity
 - Fix #64
   - Replacing the use of  "volumetric" with "volume_fraction" where appropriate
 - Resolve #82 
   - Changed all mentions of "sea level" to "mean sea level" for clarity and consistency with CF standard names
 - Finish fixing #118 
   - Changing the remaining format statements to f-strings
 - Other fixes
   - Updating github action versions per comments on #145 
   - More appropriate error types for exceptions raised in `tools/lib/xml_tools.py`
   - Remove @mwaxmonsky and @mattldawson from CODEOWNERS

## Issues
Resolves #24, #64, #82, #118


.github/workflows/pull_request_ci.yml
```diff
@@ -13,10 +13,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
```
climbfuji, line 16: Look how up to date we are!

climbfuji (commented):
I know this is just a draft, but I am happy with the changes.

mkavulich:
Thanks @climbfuji, I had forgotten a couple things in my original PR so I marked it as a draft while resolving those...the only differences between the original and the latest (ready for review) version is removing Mike W from the codeowners and changing all mentions of "sea level" to "mean sea level" (#82).

mkavulich:
@climbfuji Could you give this an official approval?

@nusbaume @svahl991 I'd appreciate a review when you get a chance, thanks!

climbfuji (approved):

tools/lib/xml_tools.py
```diff
@@ -64,26 +64,20 @@ def validate_xml_file(filename, schema_file, logger, error_on_noxmllint=False):
...
-        raise ValueError(f"validate_xml_file: Filename, '{filename}', does not exist")
-    if not os.access(filename, os.R_OK):
-        raise ValueError(f"validate_xml_file: Cannot open '{filename}'")
+        raise FileNotFoundError(f"validate_xml_file: Filename, '{filename}', does not exist")
     if not os.path.isfile(schema_file):
-        raise ValueError(f"validate_xml_file: Cannot find schema file {schema_file}")
-    if not os.access(schema_file, os.R_OK):
-        emsg = "validate_xml_file: Cannot open schema, '{}'"
-        raise ValueError(emsg.format(schema_file))
+        raise FileNotFoundError(f"validate_xml_file: Cannot find schema file {schema_file}")
```
nusbaume: I might put quotes around `{schema_file}` here in case a blank string was ever passed in by mistake:


```suggestion
        raise FileNotFoundError(f"validate_xml_file: Cannot find schema file '{schema_file}' ")
```

CODEOWNERS
```diff
@@ -3,7 +3,7 @@
 
 # These owners will be the default owners for everything in the repo.
 
-*       @cacraigucar @climbfuji @dustinswales @grantfirl @mattldawson @mkavulich @mwaxmonsky @nusbaume @peverwhee @MarekWlasak @svahl991 @ss421
+*       @cacraigucar @climbfuji @dustinswales @grantfirl @mattldawson @mkavulich @nusbaume @peverwhee @MarekWlasak @svahl991 @ss421
```
nusbaume: I would remove @mattldawson as well, as I don't believe he is with NCAR anymore.  I can look for an alternative NCAR chemistry representative in case we need one for future chemistry-related name changes/additions (which I can always add to the CODEOWNERS file in a later PR).

nusbaume (approved):
Thanks for the ping @mkavulich!  I have two optional requests, but otherwise this PR looks good to me.

svahl991 (approved):
Thanks!

CODEOWNERS
mkavulich: Yes, if you can think of people to ping I think that will be very helpful; it's been a while since a chemistry question has come up but it's good to have someone on hand.

Issue #101, opened 2025-03-21
Rules for "flag", "index", and other similar control variables


mkavulich:
We have vaguely discussed a plan for handling "flag" variables, which I've been treating as a category including all variables that are specific to model code or other logical controls rather than physical quantities. This can include a number of possible variable types, often (but not always) indicated by a prefix:

1. Boolean indicators for whether a scheme, section of code, or other option is being exercised, or condition is met. Prefixes include `flag_for_`, `do_`, `is_`
2. Integer index of a particular list, table, or other data structure controlling code execution, file read/write, or other.... Prefixes include `index_of_`, `control_for_`, `identifier_for_`

The plan is to separate out all of these "control variables" into a dedicated section, with dedicated rules setting out their naming conventions, since these variables often do not fit neatly into the schema we have developed for naming physical quantities.

mkavulich:
@grantfirl @gold2718 @climbfuji I am trying to document the history and usage of these prefixes, and I hope you can help me with some context if I'm missing it. In the current rules, only two types of control variables are described: logicals with the prefix `flag_for_`, and integers with the prefix `control_for_`. However, as documented above, the current names have a wide variety of similar variables with different names. This seems to have mostly come from PR #13 where a large number of flag variables (though not all) were changed from having a `flag_for_` prefix to either `do_`, `is_`, or `disable_`, and many control integers were likewise changed from `control_for_` to `identifier_for_`. The rules were not updated to go along with this change, but I think from the context I can figure out some rules that should be added:

| Prefix  | Type | Use case | Example |
| ------------- | ------------- | ------------- | ------------- |
| `flag_for_`  | logical | A flag indicating some state or condition is true or false | `flag_for_mpi_root` indicates whether or not the code is running on the MPI root process
| `do_` | `logical`  | A flag whose value directs some behavior | `do_chemical_tracer_diagnostics` indicates to a physics physics scheme that it should compute chemical tracer diagnostics
| `identifier_for_` | `integer`  | A parameter indicating some state or condition | `identifier_for_noah_land_surface_scheme`  an integer identifying the Noah land surface model
| `control_for_` | `integer`  | A control whose value directs some behavior | `control_for_land_surface_scheme`  an integer identifying the land surface scheme type
| `index_of_` | `integer`  | An index entry for an array | `index_of_cloud_condensation_nuclei_number_concentration_in_tracer_concentration_array`  Index describing the location of  cloud condensation nuclei number concentration in the tracer concentration array

[...]

PR #149, opened 2026-06-11
Add additional allowed prefixes for booleans

nusbaume:
## Description

This PR updates the set of allowed prefixes for booleans (e.g. variables that are of type `logical` in Fortran) to include `do` and `is` along with `flag_for`.  For example, with these rule changes the following three (made up) standard names would be allowed:

```
flag_for_deep_convection
do_deep_convective_transport
is_convective_cloud
```

whereas with the current rule set all three would need to start with `flag_for`:

```
flag_for_deep_convection
flag_for_deep_convective_transport
flag_for_convective_cloud
```

which may not be as clear.

## Issues

Addresses the "boolean indicator" discussion found in issue #101

### DIFF
--- a/StandardNamesRules.rst
+++ b/StandardNamesRules.rst
@@ -196,8 +196,8 @@ Terminology
    not valid standard name characters); e.g. ``number_density_of_plus_1_ionized_he``
 
 #. For control-oriented variables, if the variable is a Fortran logical,
-   use flag_for ``_X``. If it is any other data type, use control_for ``_X``. All flags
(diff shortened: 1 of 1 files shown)


climbfuji (commented):
Is the idea that we can use either of? Are they / should they be considered equivalent? Otherwise, we need to come up with rules when to use `do_`, `is_`, `flag_for_` (not a bad thing to have imo).

nusbaume:
Superseded by #152.

PR #152, opened 2026-07-02
Update rules for logical/control variables

### LINKED ISSUE #101: Rules for "flag", "index", and other similar control variables

mkavulich:
## Description
Update the rules for logical and control variables to distinguish what prefixes are allowed and when they should be used, based on [the proposal](https://github.com/ESCOMP/ESMStandardNames/issues/101#issuecomment-3169185309) in Issue #101. Also update a few standard names based on these rules.

This PR would supersede #149 

## Issues
Resolves #101


mkavulich:
@nusbaume When you get a chance, can you look at this and see if it's a good substitute for your #149?

climbfuji (approved):
I am ok with this. There is always going to be some ambiguity what to use, but hopefully the rules PLUS existing names in the dictionary will help.

nusbaume:
Thanks for the PR @mkavulich!  I'm happy with this replacing my previous PR. 

With regards to the standard names, was there a reason why `flag_for_` was chosen over `is_`?  I personally find `is_` to be clearer, especially when `do_` is also allowed (as one is for a given state or condition, while the other is to indicate a specific action, and both make it clear as to what a `.true.` value means).  However I could certainly be convinced otherwise if people have reasons they prefer `flag_for_`.  Thanks!

mkavulich:
@nusbaume You bring up a good point; `is_` is more analogous to `do_` in the context of the definitions I proposed in this PR, and also feels naturally less ambiguous. It would also have the advantage of making names shorter. I have updated my PR accordingly.

@climbfuji Since you already provided a review, are you okay with this change?

climbfuji:
> @nusbaume You bring up a good point; `is_` is more analogous to `do_` in the context of the definitions I proposed in this PR, and also feels naturally less ambiguous. It would also have the advantage of making names shorter. I have updated my PR accordingly.
> 
> @climbfuji Since you already provided a review, are you okay with this change?

I don't have a strong opinion on this one, happy with @nusbaume's suggestion. As long as we - at some point - start using those names in the host models!

nusbaume (approved):
Thanks all!  I'm happy with this PR now.

Issue #153, opened 2026-07-02
Possible ambiguity of `wind_from_direction_at_surface` regarding degrees


svahl991:
The name `wind_from_direction_at_surface` has units of degrees, and the description says "Direction, from north, of wind speed closest to surface". 

The `_at_surface` part of the name has ambiguities, but those are being discussed in #151, and are not the point of this issue.

The question I have here is about the orientation of the degrees. Does a wind blowing out of the south, towards the north, have a direction of zero degrees or 180 degrees? Is a west-to-east wind 90 degrees or 270? I think the answers are 180 and 270, but it seems like it's possible for there to be some confusion on this, and it would be nice to make it more explicit in the description.

svahl991:
Tagging @rajichidamb on this issue since we discussed looking into the standard that CRTM uses for this, as it appears to be opposite the standard meteorological definition.

dudhia:
Yes 180 and 270.
Also I prefer wind_direction over wind_from_direction which sounds a bit awkward. I guess they are trying to avoid ambiguity, but the meaning of wind direction is quite standard.

rajichidamb:
Talked to BenjaminTJohnson, confirmed that in CRTM the wind direction from South to North corresponds to zero degrees and is following the Oceanographic standard definition.

mkavulich:
The CF names have two definitions for wind direction (differences in *italics*, important definition in **bold**):

1. `wind_from_direction`: Wind is defined as a two-dimensional (horizontal) air velocity vector, with no vertical component. (Vertical motion in the atmosphere has the standard name upward_air_velocity.) In meteorological reports, the direction of the wind vector is usually (but not always) given as the direction from which it is blowing (wind_from_direction) (westerly, northerly, etc.). In other contexts, such as atmospheric modelling, it is often natural to give the direction in the usual manner of vectors as the heading or the direction to which it is blowing (wind_to_direction) (eastward, southward, etc.) *"from_direction" is used in the construction X_from_direction* and indicates the direction *from which the velocity vector of X is coming.*
[...]

PR #155, opened 2026-07-23
Improve definitions and usage of "surface" vs "surface_adjacent_layer" for wind variables, as well as "direction"

### LINKED ISSUE #151: Confusion about wind_speed_at_surface variants
I was wondering if we could get some clarity on the meaning of some of the wind names in the standard. The standard contains `eastward_wind_at_surface` and `eastward_wind_at_10m`. It does _not_ contain `eastward_wind_at_surface_adjacent_layer`. (Although other variables have `*at_surface_adjacent_layer` variants. ) 

The standard also contains `ratio_of_wind_at_surface_adjacent_layer_to_wind_at_10m`. Is `eastward_wind_at_surface` equivalent to `eastward_wind_at_surface_adjacent_layer`? And thus `eastward_wind_at_surface` times `ratio_of_wind_at_surface_adjacent_layer_to_wind_at_10m` is equal to `eastward_wind_at_10m`? 

If this is not correct, then what is the difference between `eastward_wind_at_surface` and `eastward_wind_at_10m`? 

We had a lot of confusion and discussion about this at a recent JEDI code sprint.

mkavulich:
## Description
As discussed in issues #151 and #153, there are shortcomings in both the names and definitions surrounding near-surface wind. This PR implements the changes discussed in those issues:

 - `wind_at_surface` variables are now `wind_at_surface_adjacent_layer`
 - We now have `wind_to_direction` in addition to `wind_from_direction`
 - `direction` is given a definition in the rules

## Issues
Resolves #151, #153

### DIFF
--- a/Metadata-standard-names.md
+++ b/Metadata-standard-names.md
@@ -601,10 +601,12 @@ Variables defining or relating to timing, dates, calendar, and related concepts
     * `real`: units = kg kg-1
 * `water_vapor_mixing_ratio_wrt_moist_air_two_timesteps_back`: Specific humidity (water vapor mass mixing ratio with respect to moist air) two timesteps back
     * `real`: units = kg kg-1
-* `wind_from_direction_at_surface`: Direction, from north, of wind speed closest to surface
+* `wind_from_direction_at_surface_adjacent_layer`: Direction (clockwise from north) wind vector is pointing away from, at the layer closest to surface.
     * `real`: units = degrees
-* `wind_speed_at_surface`: Scalar wind speed closest to surface
+* `wind_speed_at_surface_adjacent_layer`: Scalar wind speed at layer closest to surface
     * `real`: units = m s-1
+* `wind_to_direction_at_surface_adjacent_layer`: Direction (clockwise from north) wind vector is pointing towards at the layer closest to surface.
+    * `real`: units = degrees
 * `x_wind`: Horizontal wind in a direction perpendicular to y_wind
     * Equivalent CF name: `x_wind`
     * `real`: units = m s-1
--- a/Metadata-standard-names.md
+++ b/Metadata-standard-names.md
@@ -470,7 +470,7 @@ Variables defining or relating to timing, dates, calendar, and related concepts
     * `real`: units = m s-1
 * `eastward_wind_at_10m`: Wind vector component at 10 meters above surface, positive when directed eastward
     * `real`: units = m s-1
-* `eastward_wind_at_surface`: Wind vector component closest to surface, positive when directed eastward
+* `eastward_wind_at_surface_adjacent_layer`: Wind vector component at layer closest to surface, positive when directed eastward
     * `real`: units = m s-1
 * `geopotential_at_interfaces`: Geopotential at interfaces
     * `real`: units = m2 s-2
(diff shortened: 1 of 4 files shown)


jianjunj:
@mkavulich   Can you confirm that "wind_at_surface_adjacent_layer" is "wind at the lowest model layer" ?

jianjunj:
> @mkavulich Can you confirm that "wind_at_surface_adjacent_layer" is "wind at the lowest model layer" ?

I see.  It is.

svahl991 (commented):
Could we also look to see which wind `*_at_surface_adjacent_layer` names do not have a corresponding `*_at_10m` name, and add them where necessary? Then they're there if/when we need them.

svahl991:
> @mkavulich Can you confirm that "wind_at_surface_adjacent_layer" is "wind at the lowest model layer" ?

FWIW, I've encountered someone else who finds the "surface adjacent layer" wording more confusing than "lowest model layer". Personally, I don't have a problem with either one, and think both are better than the "at surface" being replaced. I don't know if I consider it worth changing, but I thought I'd mention it.
https://github.com/JCSDA-internal/ufo/issues/4313#issuecomment-5352176745

svahl991:
Tagging @ClaraDraper-NOAA @BenjaminRuston @BenjaminTJohnson  @mikecooke77

mkavulich:
@svahl991 I'm unable to view the JCSDA internal repository, so let me know if there's any specific context I'm missing. 

The `surface_adjacent_layer` wording has been in use for many years in CCPP, so I'm hesitant to change it without thorough discussion, but I won't reject the idea offhand. Expanding my usual "atmospheric thinking" into the ocean I can see how there may be some strangeness/ambiguity there, since any modeling system that incorporates information underneath the surface would have a different definition of "surface-adjacent" than the atmosphere does. And in fact it seems that we already have at least one standard name that deviates from this standard: `mass_content_of_water_in_top_soil_layer`.  I am not sure if there are any individual models that might incorporate layers both above and below the surface, but if that existed that would for sure be a deal-breaker for keeping the current wording.

That said, I can also envision scenarios where "surface-adjacent layer" is *not* synonymous with "bottom layer of atmosphere". For example, an atmospheric state interpolated to pressure coordinates can easily have "layers" that are below the physical surface (I clearly remember this scenario giving me fits in my masters thesis work!). In that case, if we did decide to change the wording, something like "atmospheric layer immediately above the surface" would be required. 

I will bring this up for wider discussion at next week's meeting.

svahl991:
To be clear, I realize changing the `surface_adjacent_layer` terminology would be a pretty big change, and I personally don't think it's worth it, and feel the wording is clear enough, and there are much bigger fish to fry.

But here's the PR comment you couldn't follow the link to. It came in the context of a discussion of making the name changes that are in this PR within JEDI code. It was similar to @jianjunj above, so I just wanted to mention it.
> To be frank, wind_speed_at_surface_adjacent_layer (unclear what "surface adjacent" means) is confusing to me as well.
Is it just the "wind_speed" at the lowest model level?
Then, will wind_speed_at_lowest_model_level be more clear?

climbfuji (approved):
I support this change based on
- `at_surface_adjacent_layer` is much clearer and more meaningful than `at_surface`
- `at_surface_adjacent_layer` already exists in the repository (x_wind, y_wind)
- `layer` and `interfaces` are already being used to discriminate between half and full levels, whereas the term `model level` has intentionally been dropped to avoid ambiguity

mkavulich:
@svahl991 I'll give you a few days to respond since I know you're on vacation, but unless you still have objections I'll go ahead and merge this per this week's discussion.

svahl991:
Before merging, can you add the corresponding `*_at_10m` variants for the variables that are missing them? I think the variables that need them are `x_wind`, `y_wind`, `wind_speed`, `wind_from_direction`, and `wind_to_direction`. I know we have need of the 10m winds for speed and direction in JEDI. I don't know if the 10m x/y winds are needed anywhere or not.

mkavulich:
@svahl991 Those names are now added, sorry I do remember now that you had requested those earlier.

climbfuji (approved):

svahl991 (approved):
Thank you!

